PR #8, opened 2016-08-30
Use minified versions of onenoteapi and picker

mttwc:
Please note that you will need to run npm install on OneNotePicker-JS and WebClipper repos after this is checked in.


leakymemory:
Looks good to me. 👍

iconix:
:ok_hand:

PR #10, opened 2016-09-01
Fix the Safari first run issue

### LINKED ISSUE #1: Safari: Page needs to be refreshed after extension install
Because in Safari we inject a listener into every page which is needed when we invoke the clipper, after you install (or upgrade) the extension any existing pages need to be refreshed before the clipper will work.

We need to detect this case and prompt the user to refresh the page before invoking the clipper.

leakymemory:
When the extension is first installed, we want to open up the post-install page on onenote.com.  In Safari, the logic was broken such that the page was not able to be loaded.  One of the side-effects of this was that the first impression of the extension was that it didn't seem to work (the page that was currently loaded didn't yet have the injected code).

Fixing the first run logic won't take care of the refresh issue, but it will certainly give a better first impression (and is inline with the other extensions at this point).

Misc notes on the change:
- I removed the inline install logic for Safari because there is no inline install.  The extension is always installed by navigating to Safari's extensions web page.
- The change in SafariExtension.ts is strictly about cleanup.  Technically it has nothing to do with the fix.

Fixes #1

### DIFF
--- a/src/scripts/extensions/extensionBase.ts
+++ b/src/scripts/extensions/extensionBase.ts
@@ -48,15 +48,17 @@ export abstract class ExtensionBase<TWorker extends ExtensionWorkerBase<TTab, TT
 		this.auth = new AuthenticationHelper(this.storage, this.logger);
 		this.tooltip = new TooltipHelper(this.storage);
 
+		let clipperFirstRun = false;
+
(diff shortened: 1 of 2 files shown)


mttwc:
Looks good! Glad this is working now :)

PR #11, opened 2016-09-01
[Bug] Extract embedded YouTube videos on KhanAcademy pages

### LINKED ISSUE #3: KhanAcademy embedded videos are not being extracted
We will likely need to add some KhanAcademy-specific client-side code to manually extract videos from KhanAcademy.

Also, it would be nice if we can have a folder of scripts that does each of these site-specific extractions, one site per file.

VishaalK:
We use a RegExp to check for the YouTube video ID in the `pageContent`, and insert an iframe with that ID.

Fixes #3


src/scripts/domParsers/videoUtils.ts
```diff
@@ -21,7 +22,7 @@ export module VideoUtils {
 		let hostname = Utils.getHostname(pageUrl);
 
 		for (let domain in SupportedVideoDomains) {
-			if (typeof(domain) === "string" && hostname.indexOf(domain.toLowerCase() + ".") > -1) {
+			if (typeof (domain) === "string" && hostname.indexOf(domain.toLowerCase() + ".") > -1) {
```
leakymemory: `typeof domain`?

src/scripts/domParsers/videoUtils.ts
```diff
@@ -65,6 +67,23 @@ export module VideoUtils {
 		return values;
 	}
 
+	export function getKhanAcademyVideoSrcValue(pageContent: string): string {
```
leakymemory: I realize that we're not consistent through the codebase, but we should use method header comments to at least keep this file looking good. :-)

Note that this is true for the other two new functions below as well.

leakymemory:
Nice job.  I'm glad to see plugging into new sites is so straight forward. 👍

src/scripts/domParsers/videoUtils.ts
```diff
@@ -6,7 +6,8 @@ export module VideoUtils {
 
 	export enum SupportedVideoDomains {
```
mttwc, line 10: Currently, all of our domain-specific video-grabbing logic lives in videoUtils, and we use subtly different ways to extract videos from each differing domain, but the end purpose is still the same. I wonder if we can break this up further into smaller classes to make things a bit simpler.

src/tests/domParsers/khanAcademyVideo_tests.ts
```diff
@@ -0,0 +1,102 @@
...
+	"<div data-youtubeid=\"video_8-5DTsl1V5k\"> </div>" // double quotations
+];
+
+let pageContentWithMultipleHyphenatedVideoIds = [
+	"<div id='video_8-5DTsl1V5k'></div> <DIV ID='video_8-53sl1V5k'></DIV> <div id='video_4ba196611'> </div>",
+];
+
+QUnit.module("videoUtils");
+
+test("videoDomainIfSupported", () => {
```
mttwc: The test name should be fleshed out so that if it fails, the person reading the console will know immediately what is going on

src/tests/domParsers/khanAcademyVideo_tests.ts
```diff
@@ -0,0 +1,102 @@
+/// <reference path="../../../typings/main/ambient/qunit/qunit.d.ts" />
+
+import {VideoUtils} from "../../scripts/domParsers/videoUtils";
```
mttwc: See above comment about splitting up the different domain-specific logics into different class (perhaps implementations of an interface). We can then split the test files up into smaller ones as well. Generally, we want files and their matching test-files to be a 1:1 ratio.

mttwc:
Some comments above. To make things simpler and more scalable, I think videoUtils should be split up into 3 classes, one for each of the domains we're supporting, and have them implement an interface of some sort.

VishaalK:
I'll see how much effort re-organization would take. I'm not sure it is worth it. The API only supports YouTube/Vimeo and these are the big few sites with those embedded right now.

src/scripts/domParsers/youtubeVideoExtractor.ts
```diff
@@ -0,0 +1,75 @@
...
+		let youTubeId;
+		if (Utils.getPathname(youTubeUrl).indexOf("/watch") === 0) {
+			youTubeId = Utils.getQueryValue(youTubeUrl, this.youTubeVideoIdQueryKey);
+			if (Utils.isNullOrUndefined(youTubeId)) {
+				return;
+			}
+		}
+
+		if (Utils.getPathname(youTubeUrl).indexOf("/embed") === 0) {
+			let youTubeIdMatch = youTubeUrl.match(/youtube\.com\/embed\/(\S+)/);
```
iconix, line 28: curious, why the switch from `\w` to `\S`?

src/scripts/domParsers/videoExtractor.ts
```diff
@@ -0,0 +1,9 @@
+export interface VideoExtractor {
+	// Our 3 Domains all have the concept of a "VideoId" embedded somewhere in the page
```
iconix, line 2: this isn't strictly true, right? YouTube.com doesn't have this concept

src/scripts/domParsers/videoExtractor.ts
```diff
@@ -0,0 +1,9 @@
...
+	getVideoIds(pageUrl: string, pageContent: string): string[];
+
+	// We can use the video ideas to generate or retrieve src values that we can embed into our article preview
+	getVideoSrcValues(pageUrl: string, pageContent: string): string[];
```
iconix, line 6: it's a little weird to me that we require passing both `pageUrl` and `pageContent` into this method, but only one or the other will get used, depending on which domain we're on... might it be better then to trust the caller to pass in the correct page info, and then just have a `pageInfo` (or similar) parameter?

src/scripts/domParsers/khanAcademyVideoExtractor.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export class KhanAcademyVideoExtractor implements VideoExtractor {
+	private youtubeExtractor: YoutubeVideoExtractor;
+
+	constructor() {
+		this.youtubeExtractor = new YoutubeVideoExtractor();
+	}
+
+	public getVideoIds(pageUrl: string, pageContent: string): string[] {
+		let regex = /id\s*=\s*("\s*video_(\S+)\s*"|'\s*video_(\S+)\s*')/gi;
```
iconix, line 15: I found this kind of bothersome to do while I was implementing Vimeo, but it will probably be useful for the future: could you add comments describing the regexes? :)

src/scripts/domParsers/khanAcademyVideoExtractor.ts
```diff
@@ -0,0 +1,50 @@
+import {VideoExtractor} from "./videoExtractor";
+import {VideoUtils} from "./videoUtils";
+import {YoutubeVideoExtractor} from "./youtubeVideoExtractor";
+
+import {Utils} from "../utils";
+
+export class KhanAcademyVideoExtractor implements VideoExtractor {
+	private youtubeExtractor: YoutubeVideoExtractor;
```
iconix, line 8: Something about this `KhanAcademyVideoExtractor` actually holding a `YoutubeVideoExtractor` makes me wonder: how about `KhanAcademyVideoExtractor extends YoutubeVideoExtractor`? 

I know it may feel kind of pointless because we'd override most of the `YoutubeVideoExtractor` method implementations... but hey, then `KhanAcademyVideoExtractor.createEmbeddedVideos` can call `super.createEmbeddedVideos` instead :D

More seriously, I also feel like using this pattern would make it quicker to grasp what's going on here

iconix:
^Unless I'm missing something, you can't merge or update reviews :( So I'll just "Approve" in the old way:

LGTM :100: Thanks for tackling this bug! I like the cleanup that's been done to make things more reusable.

src/scripts/domParsers/khanAcademyVideoExtractor.ts
```diff
@@ -0,0 +1,50 @@
...
+
+	constructor() {
+		this.youtubeExtractor = new YoutubeVideoExtractor();
+	}
+
+	public getVideoIds(pageUrl: string, pageContent: string): string[] {
```
mttwc, line 14: Thinking if we should rename pageContent to pageHtml across these methods for clarity

src/scripts/domParsers/khanAcademyVideoExtractor.ts
```diff
@@ -0,0 +1,50 @@
...
+		return VideoUtils.matchRegexFromPageContent(pageContent, regexes);
+	}
+
+	public getVideoSrcValues(pageUrl: string, pageContent: string): string[] {
+		if (Utils.isNullOrUndefined(pageContent)) {
+			return;
+		}
+
+		let videoIds = this.getVideoIds(pageUrl, pageContent);
+		if (Utils.isNullOrUndefined(videoIds)) {
```
mttwc, line 27: You could collapse these two if statements (the null/undefined check and the length check)

src/scripts/domParsers/khanAcademyVideoExtractor.ts
mttwc: Personally I prefer not using the extends and all those overrides because typescript currently doesn't support amazing support for safely overriding (i.e., it's easy for the dev to make a mistake such as forgetting to override something, or mispelling a method name). The thing I'd do would be to add a comment here saying that we use the same method as youtube to create embedded videos.

src/scripts/domParsers/vimeoVideoExtractor.ts
```diff
@@ -0,0 +1,72 @@
...
+
+import {Utils} from "../utils";
+
+export class VimeoVideoExtractor implements VideoExtractor {
+	private dataOriginalSrcAttribute = "data-original-src";
+
+	/**
+	 * Return id for a video on Vimeo.com
+	 */
+	getVideoIds(pageUrl: string, pageContent: string): string[] {
```
mttwc, line 13: Add public keywords to this file for consistency?

src/scripts/domParsers/youtubeVideoExtractor.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export class YoutubeVideoExtractor implements VideoExtractor {
+	private youTubeWatchVideoBaseUrl = "https://www.youtube.com/watch";
+	private youTubeVideoIdQueryKey = "v";
+	private dataOriginalSrcAttribute = "data-original-src";
+
+	/**
+	 * Return the ID of the video in the YouTube URL as an array
+	 */
+	getVideoIds(youTubeUrl: string, pageContent: string): string[] {
```
mttwc, line 14: Same as above

mttwc (changes requested):
Few minor comments. Nice refactor!

src/scripts/domParsers/khanAcademyVideoExtractor.ts
VishaalK: I left it as composition instead of inheritance because I don't think this follows the "is-a" rule of thumb. That being said, it doesn't really follow the "has-a" rule either :P. For inheritance, I also don't make use of a lot of the YouTube code and another good rule of thumb I've read about it is to avoid inheritance unless the vast majority of the code in the superclass is re-used.

src/scripts/domParsers/videoExtractor.ts
VishaalK: I'll correct this. Writing comments so hard -.-

src/scripts/domParsers/videoExtractor.ts
VishaalK: I considered passing in `PageInfo`. I will investigate if it easy enough to do. It would simplify the interface and make it a little more flexible.

src/scripts/domParsers/khanAcademyVideoExtractor.ts
VishaalK: Sure!

src/scripts/domParsers/vimeoVideoExtractor.ts
VishaalK: Will do!

Edit: Done!

src/scripts/domParsers/youtubeVideoExtractor.ts
VishaalK: Done!

src/scripts/domParsers/khanAcademyVideoExtractor.ts
VishaalK: Done!

src/scripts/domParsers/khanAcademyVideoExtractor.ts
VishaalK: Do you think we should pipe `pageInfo` through from `clipper.tsx` -> `augmentationHelper.ts` to all  this extraction stuff?

src/scripts/domParsers/youtubeVideoExtractor.ts
VishaalK: Some of the videos on KhanAcademy have hyphens in their data-ids, and `\w` does not cover those. Since they are also YouTube vids I changed the youtube regexes as well.

src/scripts/domParsers/videoExtractor.ts
VishaalK: I corrected it by deleting the comment

src/scripts/domParsers/khanAcademyVideoExtractor.ts
mttwc: I don't think so. I would love to keep this as seperate as possible from the clipper world (pageInfo has a reference to ONApi)

PR #13, opened 2016-09-02
Refactored storage, safer 'setValue'

### LINKED ISSUE #4: Create a StorageBase wrapper that does sanity checks for setters
For example, we should not set Notebooks in localStorage if UserInfo is not set, and we should abstract this problem away from the developer having to do this check every time.

This one will need a bit of thinking. Was originally going to write a wrapper for storageBase, however, the logic where we know if the user is signed in or not lives in the UI, whereas storageBase lives in the background. I think a possible way of solving this is having two classes that implement some interface representing storage, one of them living in the UI and one in the back. Need to think about this one more as one is synchronous (background -> itself) in nature and the other is asynchronous (UI -> background).

The class itself is a bit of a mess (e.g., getFreshValue is not very relevant to the class, storageBase is a silly name) so we could break it up a bit :)

mttwc:
Main points:

storageBase.ts has been broken up into a bunch of different single-responsibility scripts in the /http and /storage folders
- http/
  - http.ts (plain-old http, currently only implements GET as that's all we ever used from storageBase.ts)
  - cachedHttp.ts ('GET' requests with a storage object, used to cache stuff that hasn't expired yet)
  - clipperCachedHttp.ts (cachedHttp, but with Clipper-specific logging)
- storage/
  - clipperData.ts (think of this as the new storageBase.ts that encapsulates the other smaller objects)
  - clipperStorageGateStrategy.ts (StorageGateStrategy implementations encapsulate the ability to 'gate' setValue calls into storage based on a set of rules. For example, don't allow setting of notebooks if the userInfo doesn't exist)
  - clipperStorageKeys.ts (these have been moved out from constants.ts)

UTs can be found in the corresponding folders.
frontEndGlobals.ts has been refactored to use a class instead of a module. This was because I wanted 'safer' setting of the extCommunicator (i.e., initialize the storageAsync object with the communicator as soon as we set it).

Manually tested on all extension-supporting browsers so far, though I suggest we leave this sitting in master for a bit to test internally before we ship it with a version.

Fixes #4


src/scripts/storage/clipperData.ts
```diff
@@ -0,0 +1,49 @@
...
+ * sanity-check for what information gets stored locally.
+ */
+export class ClipperData implements Storage {
+	private storage: Storage;
+	private storageGateStrategy: StorageGateStrategy;
+	private cachedHttp: ClipperCachedHttp;
+
+	constructor(storage: Storage, logger?: Logger) {
+		this.storage = storage;
+		this.storageGateStrategy = new ClipperStorageGateStrategy(storage);
```
VishaalK, line 23: As I understand it, `clipperData.ts` has a single strategy. This strategy is the concrete `ClipperStorageGateStrategy` which gates 2 keys based on `userInfo`'s presence. Is there a way we can reverse the interface, so we can inject strategies into the `clipperData`? Going a step further, we can specify a strategy on a per key basis? It might not be worth it, but if we need to add another gating strategy to something in ClipperData I think it would make it easier. Also it would make testing a little easier imo (although I've looked at your tests and they are very complete).

I am envisioning injecting `clipperData` with a map of keys to strategies (`{[string]: StorageGateStrategy}` or `StrategyMap[]` where `StrategyMap = {[string]: StorageGateStrategy`) or possibly having a method to `setStrategies` for specific keys (`ClipperData::setStrategy(keys: string[], strategy: StorageGateStrategy)`).  ClipperData can check the strategy (if one exists) for the key passed into setValue.

Even if we don't end up implementing it like this, I wonder what the pros/cons of this method versus the current implementation would be.

src/tests/storage/clipperData_tests.ts
```diff
@@ -0,0 +1,121 @@
...
+		"Notebooks should not be stored if userInformation does not exist in storage");
+});
+
+test("getAndCacheFreshValue should store notebooks if it's setting notebooks and userInformation is in storage", () => {
+	let key = ClipperStorageKeys.cachedNotebooks;
+	let expectedTimeStampedData = JSON.stringify({
+		parsedResponse: "{ notebooks: {} }",
+		request: undefined
+	});
+
```
VishaalK, line 86: Commenting here for my own purposes about code design. **ignore this**

`storageGateStrategy_tests.ts` tests the concrete strategy we've implemented, and this file tests the concrete class that wraps that concrete strategy. I'm gonna think about how the other approach would look in a testing setting.

VishaalK:
Great job. Much more de-coupled, the tests are very clear. Ship it!

src/scripts/storage/clipperData.ts
mttwc: Mapping keys to strategies is an interesting idea ... I'll have a think about it. As for dependency injection, I decided to hardcode the implementation into the constructor itself as it seems like a 1:1 relationship, and if we constructor-inject it, I was thinking that we'd expose the implementation to the outside world, which is unwanted. I can definitely see the benefits of DI, which makes me wonder if we should use a DI framework some time down the road :P
I'll ponder your comments and make changes as necessary. Thanks!

src/tests/storage/clipperData_tests.ts
mttwc: Adding my 2c, the reason why I tested this the way I did was because I wanted to do so in a blackbox manner (i.e., we don't know there's a gating strategy as part of the implementation). If we use DI, more tests will need to be added for better coverage.

VishaalK: ^ yeah you are right. If the metric for good design is how testable something is, and how those tests correspond to the actual application, then the blackbox design is clearly better. It is fully black box. This is the initial reason I asked because the tests are at odds which how I think the code should be structured.

src/scripts/clipperUI/frontEndGlobals.ts
```diff
@@ -5,25 +5,55 @@ import {Communicator} from "../communicator/communicator";
...
 
-	export function getUserSessionId(): string {
-		return this.sessionId.get();
+export class Clipper {
+	private static injectCommunicator: Communicator;
+	private static extensionCommunicator: Communicator;
+
+	private static storage: StorageAsync;
+
+	// Public for convenience?
```
iconix, line 17: is this comment intended as a exploratory TODO? because then perhaps it would be better to track as an issue

src/scripts/clipperUI/frontEndGlobals.ts
```diff
@@ -5,25 +5,55 @@ import {Communicator} from "../communicator/communicator";
...
+	public static getExtensionCommunicator(): Communicator {
+		return Clipper.extensionCommunicator;
+	}
+
+	public static setExtensionCommunicator(extensionCommunicator: Communicator) {
+		Clipper.extensionCommunicator = extensionCommunicator;
+		Clipper.setUpRemoteStorage(extensionCommunicator);
+	}
+
+	public static getValue(key: string, callback: (value: string) => void): void {
```
iconix: going `Clipper.Storage`  -> `Clipper` in order to abstract away storage makes the purpose of `getValue`/`setValue` feel kind of vague... consider a name change to `getStoredValue`/`storeValue` (or something else more descriptive... not great with naming myself)

src/scripts/extensions/authenticationHelper.ts
```diff
@@ -5,25 +5,28 @@ import {SmartValue} from "../communicator/smartValue";
...
 declare var browser;
 
 export class AuthenticationHelper {
 	public user: SmartValue<UserInfo>;
 	private logger: Logger;
-	private storage: StorageBase;
+	private clipperData: ClipperData;
 
-	constructor(storage: StorageBase, logger: Logger) {
+	constructor(storage: ClipperData, logger: Logger) {
```
iconix: nit: rename parameter to `clipperData` as well for consistency

src/scripts/http/http.ts
```diff
@@ -0,0 +1,50 @@
+/// <reference path="../../../node_modules/onenoteapi/target/oneNoteApi.d.ts" />
+
+interface ResponsePackage {
+	parsedResponse: string;
+	request: XMLHttpRequest;
+}
+
+/**
+ * Helper class for performing http requests. This is a WIP and has barebones functionality, but
```
iconix: minor: a WIP?

src/scripts/storage/clipperData.ts
```diff
@@ -0,0 +1,49 @@
...
+		// We pass 'this' as the Storage object as it handles all the sanity-check gating in the setValue
+		this.cachedHttp = new ClipperCachedHttp(this, logger);
+	}
+
+	public setLogger(logger: Logger) {
+		this.cachedHttp.setLogger(logger);
+	}
+
+	public getAndCacheFreshValue(key: string, getRemoteValue: GetResponseAsync, updateInterval?: number): Promise<TimeStampedData> {
```
iconix: might just be me, but saying get _and_ cache feels like this is now promising to make the network call (get) so that it can follow through on its new promise to cache, which isn't what's happening here. a get will always happen of some sort, but it's possible the caching will not if the value was already fresh.

did others find the old `getFreshValue` name confusing? perhaps it's familiarity, but I was fine with that

src/scripts/http/cachedHttp.ts
```diff
@@ -0,0 +1,121 @@
...
+
+			this.getAndCacheRemoteValue(key, getRemoteValue).then((timeStampedData) => {
+				resolve(timeStampedData);
+			}, (error) => {
+				reject(error);
+			});
+		});
+	}
+
+	public getAndCacheRemoteValue(key: string, getRemoteValue: GetResponseAsync): Promise<TimeStampedData> {
```
iconix: I wish there was a middle ground between `public` and `private` in TS... because this implementation (and the one in `ClipperCachedHttp`) aren't implementations we actually want to be using in our codebase, I imagine. So having them `public`... feels too tempting/easy for someone to lose track of some of the abstraction going on here and just start calling it (in all of its log-less, throwing glory ;P)

I get the pattern you're going for, but could we perhaps just make some of these low-level implementations (that are currently only being directly used in tests) throw an Not Implemented error instead and then condense all desired implementation behavior into `ClipperData`?

src/scripts/http/http.ts
mttwc: Currently, we only implement GET here because it's the only thing we use from what I can see. I'll double check this.

src/scripts/storage/clipperData.ts
mttwc: You convinced me :P
I renamed it back to getFreshValue. Resolved

src/scripts/http/cachedHttp.ts
mttwc: I'm not sure if we call this explicitly from the outside, but I do believe there are times where we actually force a fresh get of remote values (by setting updateInterval to 0). I may have forgotten to convert those calls to this new method call, so I'll do a double check :)

src/scripts/clipperUI/frontEndGlobals.ts
mttwc: Good idea. I left this as public static because this file used to be a module rather than a class. I wanted to keep the changes as minimal as possible so it's not a nightmare to review, so I wanted to keep the get/set syntax the same. I'll add an Issue.

src/scripts/clipperUI/frontEndGlobals.ts
mttwc: Good idea!

mttwc: Resolved

src/scripts/extensions/authenticationHelper.ts
mttwc: Resolved

src/scripts/http/cachedHttp.ts
mttwc: I made the method protected and am changing the UTs to now hit the public method (with an updateInterval of 0). Resolved

mttwc:
@iconix I've added getValues to our storage classes/interfaces which you can use in your changes :)

PR #15, opened 2016-09-02
Reimplement Firefox onInstall logic

### LINKED ISSUE #14: (Firefox) Port the onInstall logic in Safari to Firefox

mttwc:
Firefox's implementation of WebExtensions currently does not include onInstall() (i.e., the method is undefined). We have some logic for FRE where if the user does not have a clipperId, we assume it's a new user. WebExtensions does not use this as we use onInstall instead (which does not work for Firefox). Ported Firefox onInstall logic back to this old way of doing it. Fixes #14


leakymemory:
Looks good.

PR #18, opened 2016-09-06
Add a couple more elements to remove before sending the DOM

### LINKED ISSUE #17: Some articles are missing images.
Here is an example page:
http://www.cooksillustrated.com/recipes/9065-grilled-shrimp-and-vegetable-kebabs

It looks like there are a couple of issues on this site:

1) The full page screenshot is off a bit.  It looks like we're garbling up the layout at the top.
2) In article mode, the image is missing.

For #2, the reason is because the <picture> element contains some child <source> elements which need to be deleted.  Also, there is a transparent SVG just below that needs to be removed.

leakymemory:
Add a couple more elements to remove before sending the DOM to the article extractor.  Specifically, remove the svg and source elements.

Fixes #17

### DIFF
--- a/src/scripts/domParsers/domUtils.ts
+++ b/src/scripts/domParsers/domUtils.ts
@@ -33,6 +33,7 @@ export module DomUtils {
 		export const noscript = "noscript";
 		export const progress = "progress";
 		export const script = "script";
+		export const source = "source";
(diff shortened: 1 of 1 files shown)


mttwc:
Looks good! :sheep:it

PR #21, opened 2016-09-08
Add ONApi to the building of the pagenav scripts

### LINKED ISSUE #20: OneNoteApi is not defined in chromePageNavInject.js
This is likely happening on our other page nav inject scripts. This does not seem to affect the operation of the tooltip, so it may be possible that we're referencing the Api somewhere in the script despite us not needing to (I don't see why we should be doing so in the first place). We either need to remove the reference to OneNoteApi or ensure it gets built into the script during build.

mttwc:
Currently, we throw an error when a page nav tooltip appears as ONApi is not included in the script when we try and reference it. This causes some of our logging to be thrown off as a result, but luckily does not affect the user experience. This will fix that.

Fixes #20

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -611,6 +611,7 @@ function exportChromeJS() {
 	]).pipe(concat("chromeInject.js")).pipe(gulp.dest(targetDir));
 
 	var chromePageNavInjectTask = gulp.src([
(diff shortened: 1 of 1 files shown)


leakymemory:
Looks good.

VishaalK:
:star:x5

PR #22, opened 2016-09-12
[Feature] Ratings prompt

iconix:
This feature enables the Web Clipper to receive positive and negative feedback from "active" users.

When a user's engagement with the Web Clipper qualifies them, we will show an additional sub-panel after a successful clip, which asks: "Enjoying the Web Clipper?" We will then log their response ("Yes, it's great!" vs "Not really...") depending on button clicked.

Additionally, if rate and feedback URLs are provided in Web Clipper settings, we will show the user additional prompts that will direct them to places where they can leave further feedback.

The "when to show" logic for this ratings sub-panel is where things get a little complicated. In broad strokes, the checks break down as follows:
1. Is this feature enabled for the client being used (e.g., ChromeExtension, EdgeExtension, Bookmarklet, etc.)?
2. Is the user "active" (dependent on # successful clips)?
3. Have we shown this ratings panel to the user before? If so, what kind of response did we get (no response, positive response, negative response)?

_Please see method descriptions in `RatingsHelper.ts` for further breakdown of the "when to show" logic._

**PENDING TO-DOs as of initial PR publish:**
1. Animations, as spec'd by the design team
2. Fix commented-out tests in `ratingsHelper_tests.tsx`
3. Finalize logic on 1) how to handle users who never interact with the ratings prompt, and 2) how to determine enough time has passed after negative feedback to re-prompt users? (_email discussion ongoing_)


iconix:
**This is a temporary comment with tips on how to test the logic in this PR.**

I have pushed a [`test/ratings-prompt` branch](https://github.com/OneNoteDev/WebClipper/tree/test/ratings-prompt) that contains all the changes in this PR + a few development-only changes I used to speed up my dev loop. This is what it will do for you:
1. You won't have to actually clip to see the success panel - it will mimic a clip and always be the first panel that shows up.
2. If you say "Not really..." to "Enjoying the Web Clipper?", you'll only have to wait 30 seconds (as opposed to 10 weeks) before getting re-prompted (assuming you deleted `lastBadRatingVersion` in storage as well).

If you pull the `feature/ratings-prompt-settings` branch in _Internal, it will provide the feature enabled setting for each client (note that only in Chrome is this feature turned on right now), as well as rate and feedback URLs. The availability of these settings will determine if and how the ratings sub-panel shows up!

If you're playing around with the feature and the ratings sub-panel stops showing up when you think it should be, check the "when to show" logic and try deleting one of the following storage values: `doNotPromptRatings`, `lastBadRatingDate`, `lastBadRatingVersion`, `lastSeenVersion`, `numClipSuccess`.

iconix:
**This is a temporary comment with tips on how to test the logic in this PR.**

---
I have pushed a [`test/ratings-prompt` branch](https://github.com/OneNoteDev/WebClipper/tree/test/ratings-prompt) that contains all the changes in this PR + a few development-only changes I used to speed up my dev loop. This is what testing on it will do for you:

1. You won't have to actually clip to see the success panel - it will mimic a clip and always be the first panel that shows up.
2. If you say "Not really..." to "Enjoying the Web Clipper?", you'll only have to wait 30 seconds (as opposed to 10 weeks) before getting re-prompted (assuming you delete `lastBadRatingVersion` in storage as well).

---
If you pull the `feature/ratings-prompt-settings` branch in [_Internal](https://github.com/OneNoteDev/WebClipper_Internal/pull/51), it will provide the feature enabled setting for each client (note that only in Chrome is this feature turned on right now), as well as rate and feedback URLs. **The availability of these settings will determine if and how the ratings sub-panel shows up!**

---
If you're playing around with the feature and the ratings sub-panel stops showing up when you think it should be, check the "when to show" logic and try deleting/updating one of the following storage values: `doNotPromptRatings`, `lastBadRatingDate`, `lastBadRatingVersion`, `lastSeenVersion`, `numClipSuccess`.

Thanks!

src/scripts/constants.ts
```diff
@@ -339,10 +352,14 @@ export module Constants {
 	}
 
 	export module Settings {
-		export var timeBetweenTooltips = 1000 * 60 * 60 * 24 * 7 * 3; // 21 days
-		export var noOpTrackerTimeoutDuration = 20 * 1000; // 20 seconds
+		export var fontSizeStep = 2;
+		export var maxClipSuccessForRatingsPrompt = 12; // TODO waiting for consensus on this value
+		export var maximumJSTimeValue = 8640000000000000; // http://ecma-international.org/ecma-262/5.1/#sec-15.9.1.1
```
VishaalK: can you make this multiplied out like the others?

src/scripts/clipperUI/ratingsHelper.ts
```diff
@@ -0,0 +1,538 @@
...
+	}
+
+	/**
+	 * Based on the three-part version number convention "X.Y.Z", where:
+	 *   * X = major version
+	 *   * Y = minor version
+	 *   * Z = patch
+	 * Returns the major version, X
+	 */
+	private static getMajorVersion(versionNum: string): number {
```
VishaalK: Matthew wrote `version.ts` which is a class that might make this code easier. 

I see later in the code you are checking the major versions of lastSeenVersion and badRatingMajor. It might be worth abstracting that into a function `isMajorVersionLessThanOrEqualTo` or something along those lines and adding it to `version.ts`.

src/scripts/clipperUI/clipper.tsx
```diff
@@ -620,9 +621,22 @@ class ClipperClass extends ComponentBase<ClipperState, {}> {
 
 		this.state.setState({ oneNoteApiResult: { status: Status.InProgress } });
 		SaveToOneNote.startClip(this.state).then((startClipPackage: StartClipPackage) => {
-			clipEvent.setCustomProperty(Log.PropertyName.Custom.CorrelationId, startClipPackage.responsePackage.request.getResponseHeader(Constants.HeaderValues.correlationId));
-			clipEvent.setCustomProperty(Log.PropertyName.Custom.AnnotationAdded, startClipPackage.annotationAdded);
-			this.state.setState({ oneNoteApiResult: { data: startClipPackage.responsePackage.parsedResponse, status: Status.Succeeded } });
+			RatingsHelper.incrementClipSuccessCount().then(() => {
+				// do nothing special on incrementClipSuccessCount resolve
+			}).then(() => {
+				// will happen regardless of success/failure of incrementClipSuccessCount
```
mttwc: Shouldn't incrementClipSuccessCount never reject?

src/scripts/clipperUI/clipper.tsx
```diff
@@ -620,9 +621,22 @@ class ClipperClass extends ComponentBase<ClipperState, {}> {
...
+				RatingsHelper.shouldShowRatingsPrompt(this.state).then((shouldShowRatingsPrompt) => {
```
mttwc: I'm still contemplating if it's possible to front-load getting all the info we need to determine if the ratings prompt should be shown ... please correct me if I'm wrong, but this is determined using values in storage as well as current state, so we could grab those storage values at initialization, then do this computation in the front end so we stay synchronous. Again, I haven't dove in super deep yet, but just my 2c. Will think about this more when I dive in more.

src/scripts/constants.ts
```diff
@@ -255,14 +264,18 @@ export module Constants {
...
+		export var lastBadRatingDate = "lastBadRatingDate";
+		export var lastBadRatingVersion = "lastBadRatingVersion";
 		export var lastClippedDate = "lastClippedDate";
 		export var lastSeenVersion = "lastSeenVersion";
 		export var lastInvokedDate = "lastInvokedDate";
 		export var lastSeenTooltipTimeBase = "lastSeenTooltipTime";
 		export var lastClippedTooltipTimeBase = "lastClippedTooltipTime";
 		export var locale = "locale";
 		export var locStrings = "locStrings";
+		export var numClipSuccess = "numClipSuccess";
```
VishaalK: nit and also tedious :stuck_out_tongue:, I feel like `numSuccessfulClips` is a more descriptive name.

src/scripts/clipperUI/panels/dialogPanel.tsx
```diff
@@ -15,42 +17,55 @@ export interface DialogPanelProps {
 	content: HTMLElement[];
 	buttons: DialogButton[];
 	fontFamily?: Localization.FontFamily;
+	buttonFontFamily?: Localization.FontFamily;
+	divId?: string;
```
mttwc: Consider renaming this to containerId or similar, as it's unclear which div this refers to

src/scripts/clipperUI/panels/successPanel.tsx
```diff
@@ -9,12 +9,27 @@ import {ClipMode} from "../clipMode";
...
 import {Status} from "../status";
 
 import {SpriteAnimation} from "../components/spriteAnimation";
 
-class SuccessPanelClass extends ComponentBase<{ }, ClipperStateProp> {
-	onLaunchOneNoteButton() {
+import {DialogButton, DialogPanel} from "./dialogPanel";
+
+interface SuccessPanelState {
+	currentRatingsPromptStage?: RatingsPromptStage;
```
mttwc: I think the success panel itself does not interact with the ratings prompt portion. If so, we could keep those two completely decoupled (i.e., they don't know about each other). Then maincontroller can determine if the ratings prompt should be shown, and if so, show both success panel and ratings panel. 

There's a couple of benefits I can think of for that approach:

a) We can chain animations easier as it's handled in a common parent container that is aware of all animation strategies.
b) Greater panel reuse

src/scripts/clipperUI/panels/successPanel.tsx
```diff
@@ -9,12 +9,27 @@ import {ClipMode} from "../clipMode";
...
+	userSelectedRatingsPromptStage?: RatingsPromptStage;
+}
+
+export class SuccessPanelClass extends ComponentBase<SuccessPanelState, ClipperStateProp> {
+	getInitialState(): SuccessPanelState {
+		return {
+			currentRatingsPromptStage: RatingsPromptStage.INIT,
```
mttwc: We should standardize how we name our enum values throughout the codebase.

src/scripts/clipperUI/panels/successPanel.tsx
```diff
@@ -25,24 +40,47 @@ class SuccessPanelClass extends ComponentBase<{ }, ClipperStateProp> {
...
+	private showRatingsPrompt(): any[] {
+		if (this.props.clipperState.shouldShowRatingsPrompt) {
+			let message: string = RatingsHelper.getMessage(this.state.userSelectedRatingsPromptStage);
+			let buttons: DialogButton[] = RatingsHelper.getDialogButtons(this);
+
+			if (!Utils.isNullOrUndefined(message)) {
+				let animationStrategy = RatingsHelper.getAnimationStategy(this);
+
+				return (
+					<DialogPanel
```
mttwc: We could create a RatingsPanelClass that renders a DialogPanel to make things a bit cleaner.

src/scripts/clipperUI/ratingsHelper.ts
```diff
@@ -0,0 +1,538 @@
+import {Localization} from "../localization/localization";
```
mttwc: See previous comment, if we created a RatingsPanelClass, a lot of these methods could live there instead.

src/scripts/clipperUI/panels/successPanel.tsx
VishaalK: This makes sense. In other words, Instead of the ratings prompt being a sub component of the successPanel, the components are separate and just shown on the same panel? An analogous part of the app would be how the modeButtons are handled

src/scripts/clipperUI/ratingsHelper.ts
```diff
@@ -0,0 +1,538 @@
...
+										});
+									}
+								});
+							});
+						}
+					});
+				break;
+			case RatingsPromptStage.RATE:
+				let rateUrl: string = RatingsHelper.getRateUrlIfExists(clientType);
+				if (!Utils.isNullOrUndefined(rateUrl) && rateUrl.length > 0) {
```
mttwc: This can be simplified to if (rateUrl)

src/scripts/clipperUI/ratingsHelper.ts
```diff
@@ -0,0 +1,538 @@
...
+	public static getDialogButtons(panel: SuccessPanelClass): DialogButton[] {
+		let stage: RatingsPromptStage = panel.state.userSelectedRatingsPromptStage;
+		let clipperState: ClipperState = panel.props.clipperState;
+		let clientType: ClientType = clipperState.clientInfo.clipperType;
+
+		let buttons: DialogButton[] = [];
+
+		switch (stage) {
+			case RatingsPromptStage.INIT:
+				buttons.push({
```
mttwc: Probably good to refactor each of these cases into its own method for readability.

src/scripts/clipperUI/ratingsHelper.ts
```diff
@@ -0,0 +1,538 @@
...
+				return resolve(clipperState.shouldShowRatingsPrompt);
+			}
+
+			let ratingsPromptEnabled: boolean = RatingsHelper.ratingsPromptEnabledForClient(clipperState.clientInfo.clipperType);
+			logEventInfo.ratingsPromptEnabledForClient = ratingsPromptEnabled;
+			if (!ratingsPromptEnabled) {
+				return resolve(false);
+			}
+
+			Clipper.Storage.getValue(Constants.StorageKeys.doNotPromptRatings, (doNotPromptRatingsStr) => {
```
mttwc: Looks like a good opportunity to implement some sort of getValues(..) method where you can retrieve a bunch of values all at once :)

src/scripts/clipperUI/clipper.tsx
VishaalK: Do you know if there are other places in the code where decide behavior based on localStorage data? We front-load things like `userInfo`, which pulls from localStorage (e.g. `flightingInfo`). We could use a similar architecture for this. I think it would simplify the code by reducing the number of promises.

src/scripts/clipperUI/panels/successPanel.tsx
```diff
@@ -25,24 +40,47 @@ class SuccessPanelClass extends ComponentBase<{ }, ClipperStateProp> {
 		}
 	}
 
+	private showRatingsPrompt(): any[] {
+		if (this.props.clipperState.shouldShowRatingsPrompt) {
```
VishaalK: I think you can flatten the logic in this function.

```
if (!this.props.clipperState.shouldShowRatingsPrompt) {
    return;
}

// rest of stuff. 
```

src/scripts/clipperUI/ratingsHelper.ts
VishaalK: +1  

There are other places this could be used. For example, getting the cachedNotebooks and curSectionPath in SectionPicker.tsx. I also feel like there are other places in the codebase.

src/scripts/clipperUI/ratingsHelper.ts
```diff
@@ -0,0 +1,538 @@
...
+	/**
+	 * We will show the ratings prompt if ALL of the below applies:
+	 *   * Ratings prompt is enabled for the ClientType/ClipperType
+	 *   * If StorageKeys.doNotPromptRatings is not "true"
+	 *   * If RatingsHelper.badRatingTimingDelayIsOver(...) returns true when provided StorageKeys.lastBadRatingDate
+	 *   * If RatingsHelper.badRatingVersionDelayIsOver(...) returns true when provided StorageKeys.lastBadRatingVersion and StorageKeys.lastSeenVersion
+	 *   * If RatingsHelper.clipSuccessDelayIsOver(...) returns true when provided StorageKeys.numClipSuccess
+	 */
+	public static shouldShowRatingsPrompt(clipperState: ClipperState): Promise<boolean> {
+		return new Promise<boolean>((loggingResolve, loggingReject) => {
```
mttwc: Might be good to add a comment here saying this method only wraps the internal method with logging

src/scripts/clipperUI/ratingsHelper.ts
```diff
@@ -0,0 +1,538 @@
...
+			});
+		});
+	}
+
+	/**
+	 * Get the animation strategy for the ratings subpanel of the success panel provided
+	 */
+	public static getAnimationStategy(panel: SuccessPanelClass): AnimationStrategy {
+		return new SlideFromRightAnimationStrategy({
+			extShouldAnimateIn: () => {	return panel.state.userSelectedRatingsPromptStage !== panel.state.currentRatingsPromptStage; },
```
mttwc: nit: use space, not tab

VishaalK:
I feel like `RatingsHelper.ts` is quite large and could use some breaking up, not for modularity but readability. Some preliminary thoughts including separating out the strategy for determining when we show the ratings prompt into its own object, since there are some other utility methods in `RatingsHelper.ts` like the ones about rateUrl that seem orthogonal. I think Matthew's suggestion of breaking up the `case` statements into their own functions is good. There is a pseudo state machine in there and I think those functions could live in their own file. 

I will come back to it to see if I can make some more concrete and constructive comments about how to go about doing that.  I am also waiting to make these comments since there are a number of good suggestions here and based on the next iteration it might not be necessary.

Looking forward to the :star: :star: :star: :star: :star:

src/scripts/clipperUI/ratingsHelper.ts
```diff
@@ -0,0 +1,538 @@
...
+		return date >= minimumTimeValue && date <= Constants.Settings.maximumJSTimeValue;
+	}
+
+	private static setDoNotPromptStatus(): void {
+		Clipper.Storage.setValue(Constants.StorageKeys.doNotPromptRatings, "true");
+
+		Clipper.logger.logEvent(new Log.Event.BaseEvent(Log.Event.Label.SetDoNotPromptRatings));
+	}
+
+	private static versionHasCorrectFormat(version: string): boolean {
```
mttwc: As above with @VishaalK's comment, I think this logic already exists in version.ts, if not it can go in that file, or in another file in the versioning folder as appropriate.

src/scripts/clipperUI/ratingsHelper.ts
```diff
@@ -0,0 +1,538 @@
...
+			onAfterAnimateIn: () => { panel.setState({ currentRatingsPromptStage: panel.state.userSelectedRatingsPromptStage }); }
+		});
+	}
+
+	/**
+	 * Adds 1 to the value in StorageKeys.numClipSuccess.
+	 * If the value does not exist yet or is invalid, we (re)initialize it to 1.
+	 */
+	public static incrementClipSuccessCount(): Promise<void> {
+		return new Promise<void>((resolve, reject) => {
```
VishaalK: This function and behavior could belong in `saveToOneNote::startClip(...)`? I remember implementing something like this when I did tooltips and I think it looked identical LOL (but I ended up scrapping it :stuck_out_tongue_closed_eyes:

src/scripts/clipperUI/ratingsHelper.ts
```diff
@@ -0,0 +1,538 @@
...
+								if (!Utils.isNullOrUndefined(doNotPromptRatingsStr) && doNotPromptRatingsStr.toLowerCase() === "true") {
+									logEventInfo.doNotPromptRatings = true;
+									return resolve(false);
+								}
+
+								let lastBadRatingDate: number = parseInt(lastBadRatingDateAsStr, 10);
+								let numClips: number = parseInt(numClipsAsStr, 10);
+
+								/* tslint:disable:no-null-keyword */
+									// null is the value storage gives back; also, setting to undefined will keep this kvp from being logged at all
```
mttwc: Is this extra identation intentional?

src/scripts/settings.ts
```diff
@@ -14,4 +14,13 @@ export module Settings {
 
 		return localResult.Value;
 	}
+
+	export function setSettingsJsonForTesting(jsonObject?: {}): void {
```
mttwc, line 18: Probably should go in another change, but settings.ts should not expose a method like this where it can potentially get misused. If we want to test with custom settings, an alternative approach might be to use dependency injection, where in our test environments we can inject a different object, and in prod environments, we can inject the one in the settings file.

src/styles/clipper.less
```diff
@@ -119,7 +119,6 @@
 	top: @ClipperUITopRightOffset;
 	right: @ClipperUITopRightOffset;
 	width: 100px;
-	background: @OneNotePurple;
```
mttwc, line 122: Just wondering about the lines you removed in the less files. Was there something you found that necessitated these changes?

src/scripts/constants.ts
iconix: sorry, what would the multiplication you're suggesting here be? e.g., do you find `100,000,000 (days) * 86,400,000 (ms)` more readable (as described by the spec linked to in the comment)?

src/scripts/clipperUI/ratingsHelper.ts
iconix: ah, wasn't aware of `version.ts` - thanks!

#Pending -> #Resolved

src/scripts/clipperUI/clipper.tsx
iconix: yep, which is why I don't have a reject handler. but if something unexpectedly throws inside of the method (granted, unlikely given the method), it will reject. so I left this empty resolve handler around just to ensure that the next `.then` _definitely_ always fires (since it's pretty important). 

maybe there's a cleaner way of ensuring this. reviewing the other comments though, I can see this going away altogether.

does that answer what I think you're asking?

src/scripts/clipperUI/clipper.tsx
iconix: If what I'm about to say actually does sound similar to something the Clipper already does, then please let me know.

But I think the key diff between this feature's storage dependencies and the way the rest of the Clipper currently seems to depend on storage is that this feature has a few `setValue` calls sprinkled around that 1) only happen on user interaction, and 2) the `getValue` calls need to wait for.

so for this particular code fragment, the thing keeping it async is that we need to wait for the `setValue(numClipSuccess)` to occur before doing the `getValue(numClipSuccess)` to determine if we want to show the ratings prompt. and the `setValue` only happens if the user decides to clip (and we don't fail them).

a similar need-to-wait scenario happens if the user clicks the "Not really..." button (e.g., the thing I'm having trouble unit testing right now).

@VishaalK, I had a TODO at one point that was like "do we have to chain all of these promises??" :) so what you're suggesting is using a smart value for any value being set in storage by this feature that is later required by a get from storage for this feature, yeah? that could work in terms of getting rid of some of the chaining, but one issue I see off the bat with that is making sure the `setState`s happen in the correct order after teasing this all apart. we want `shouldShowRatingsPrompt` to be set appropriately before `oneNoteApiResult.Status` is set to Succeeded (due to how mainController works).

I'm going to need to think a lot more about this... :)

#LotsPending

src/scripts/constants.ts
iconix: I think that was the original value I had but then decided to value brevity over descriptiveness. Guess this could go back

#Pending -> #Resolved

src/scripts/clipperUI/panels/dialogPanel.tsx
iconix: #Pending -> #Resolved

src/scripts/clipperUI/panels/successPanel.tsx
iconix: I originally wanted a standalone RatingsPanel component too, but when I was thinking things through at the beginning, the MainController seemed like it would inherit too much ratings logic that it didn't need to know about in order to do so.

Now that the code is written, however, that result of that logic is available in the clipper state. So it's more obvious to me now that I can move that decision to the main controller

#Pending -> #Resolved

src/scripts/clipperUI/ratingsHelper.ts
iconix: That simplification won't include the empty string case :)

#Closed

src/scripts/clipperUI/ratingsHelper.ts
iconix: I could refactor this code forever

#Pending

src/scripts/clipperUI/ratingsHelper.ts
iconix: yeah, was kind of hoping the storage refactor PR (that I still haven't looked at, sorry) might do this :P

but I could do it here, sure

#Pending -> #WontFix (in this PR - refactor means not needing to nest these calls anymore for this feature. however, @mttwc did make `getValues` available in https://github.com/OneNoteDev/WebClipper/pull/13)

src/scripts/clipperUI/panels/successPanel.tsx
iconix: #Pending -> #Resolved

src/scripts/clipperUI/ratingsHelper.ts
iconix: I'm not sure about this. I view public method comments like this as a description for the user/caller of this method (especially since when you hover over a method in VS Code etc., this is the description provided), and so the caller need not know about implementation details like that.

#WontFix for now

src/scripts/clipperUI/ratingsHelper.ts
iconix: #Pending -> Resolved

src/scripts/clipperUI/ratingsHelper.ts
iconix: yeah, I like this idea

#Pending -> Resolved

src/scripts/clipperUI/ratingsHelper.ts
iconix: yeah... because it's a comment explaining a comment :P

#Closed

src/scripts/settings.ts
iconix: Also felt a little uncomfortable with this change, hence the attempt to name it explicitly for testing only.

I can open a bug/issue to make this DI change you suggested in another branch

#Pending -> #Closed (opened https://github.com/OneNoteDev/WebClipper/issues/35)

src/styles/clipper.less
iconix: this remove in particular was because it was being overridden by the line directly below it :P

since the majority of the other removes only affect the `successPanel` (which has been restyled by design), I assume your other concern is around the removal of
`.height-animator {
    overflow: hidden;       
}`
I can try putting it back in to see if it still breaks the ratings panel render

#Pending -> #Resolved (reverted the `.height-animator` change)

src/scripts/clipperUI/panels/successPanel.tsx
iconix: You've said this once before on another PR, but to be honest I don't know what you mean by it. Want to open a bug/issue describing this task? :)

#Pending

src/scripts/clipperUI/clipper.tsx
mttwc: I was thinking that since we don't expect the function to throw, we don't need this sort of logic in there, since our UTs don't indicate any sort of throwing behaviour.

src/scripts/clipperUI/panels/successPanel.tsx
mttwc: We have other places in our codebase where we declare enums as PascalCase e.g., AnimationState, ClipMode, AugmentationModel, as opposed to entirely uppercase. We should follow the same naming convention throughout the code base :)

src/scripts/clipperUI/ratingsHelper.ts
mttwc: Empty string is equivalent to false in js, so I usually use if(myString) to mean "if myString is a non-empty string"

src/scripts/clipperUI/clipper.tsx
iconix: Yeah, I just want to be extra careful that
`this.state.setState({ oneNoteApiResult: { data: startClipPackage.responsePackage.parsedResponse, status: Status.Succeeded } });`
isn't ever not set because we (for example) tried to read a property off an undefined value inside the promise, or something - that's now with this implementation, but also if anyone updates this logical flow later on.

The panel will get stuck on loading if we ever fail to set that state

src/scripts/clipperUI/ratingsHelper.ts
iconix: oh wait, I just realized I misinterpreteted what I tested in the console - yeah you're right

#Closed -> #Pending -> #Resolved

src/scripts/constants.ts
VishaalK: Eh I guess it doesn't matter since its a max time value.

100000000 \* 24 \* 60 \* 60 \* 1000; // 100K days in milliseconds, http://ecma-international.org/ecma-262/5.1/#sec-15.9.1.1

iconix: hm, ok I can do that!

#Pending -> #Resolved

src/scripts/clipperUI/clipper.tsx
iconix: With the refactor, this flow no longer exists

#Closed

src/scripts/clipperUI/clipper.tsx
iconix: I think the recent refactor has resolved all concerns raised here, but please let me know if I've missed anything!

#Resolved

src/scripts/clipperUI/panels/successPanel.tsx
iconix: Opened https://github.com/OneNoteDev/WebClipper/issues/34

#Resolved

src/tests/settings_tests.ts
```diff
@@ -52,10 +52,11 @@ test("helper function checkValueAndDescriptionOnObject should assert when given
 	ok(checkValueAndDescriptionOnObject(valid));
 });
 
-test("All entries in setting.json should have a non-empty Value and Description", () => {
+// TODO is a non-empty check necessary?
+/*test("All entries in setting.json should have a non-empty Value and Description", () => {
```
VishaalK: Commented out test?

src/tests/clipperUI/ratingsHelper_tests.tsx
```diff
@@ -0,0 +1,596 @@
...
+import {Constants} from "../../scripts/constants";
+import {ClientType} from "../../scripts/clientType";
+import {Settings} from "../../scripts/settings";
+import {Utils} from "../../scripts/utils";
+
+import {HelperFunctions} from "../helperFunctions";
+
+// MOCK STORAGE
+
+let mockStorage: { [key: string]: string };
```
VishaalK, line 19: We have 3-ish versions of `mockStorage` out there. I see one in `tooltipHelper_tests.ts`, another in `sectionPicker_tests.ts`. Can you add a low pri PBI to unify these? Not necessary for this PR (may never be necessary 😜).

src/scripts/clipperUI/ratingsHelper.ts
```diff
@@ -0,0 +1,294 @@
...
+
+	/**
+	 * Returns true if ALL of the below applies:
+	 *   * Number of successful clips >= {Constants.Settings.minClipSuccessForRatingsPrompt}
+	 *   * Number of successful clips <= {Constants.Settings.maxClipSuccessForRatingsPrompt}
+	 *
+	 * Public for testing
+	 */
+	public static clipSuccessDelayIsOver(numClips: number): boolean {
+		// TODO MVP+? when # of successful clips > nMax, collapse panel into a Rate Us hyperlink in the footer that is always available
```
VishaalK: similar to another comment. Create a task for this somewhere

src/tests/clipperUI/ratingsHelper_tests.tsx
```diff
@@ -0,0 +1,596 @@
...
+	Clipper.Storage.setValue(Constants.StorageKeys.lastBadRatingVersion, "3.0.9");
+	Clipper.Storage.setValue(Constants.StorageKeys.lastSeenVersion, "3.0.999");
+
+	let clipperState = HelperFunctions.getMockClipperState();
+
+	let shouldShowRatingsPrompt: boolean = RatingsHelper.shouldShowRatingsPrompt(clipperState);
+	strictEqual(shouldShowRatingsPrompt, false);
+});
+
+// test("", () => { });
```
VishaalK, line 534: Tests for RatingsHelper looks very thorough !

VishaalK (approved):

src/scripts/clipperUI/panels/ratingsPanel.tsx
```diff
@@ -0,0 +1,211 @@
...
+						}
+					});
+				} else {
+					// this shouldn't happen
+					panel.setState({
+						userSelectedRatingsPromptStage: RatingsPromptStage.None
+					});
+				}
+				break;
+			case RatingsPromptStage.Feedback:
```
VishaalK, line 152: Can you split these into functions for each state for readability? The logic looks good though.

src/tests/settings_tests.ts
iconix: Yeah, wasn't sure if there was a technical reason why we disallow empty values... because now I'd like to start using empty values. I'll assume it's all good and delete this test.

#Pending -> #Resolved

src/scripts/clipperUI/frontEndGlobals.ts
```diff
@@ -16,13 +16,31 @@ export module Clipper {
 	}
 
 	export module Storage {
-		export function getValue(key: string, callback: (value: string) => void): void {
+		let storageCache: { [key: string]: string } = {};
+
+		export function preCacheValues(storageKeys: string[]): void {
+			for (let key of storageKeys) {
+				Clipper.Storage.getValue(key, () => { }, true);
```
VishaalK: I know Matthew made a `getValues` but you'd have to add in the caching in again that you already added to `getValue` (if we want to switch to that function). Just a heads up depending on who merges first.

src/scripts/clipperUI/ratingsHelper.ts
```diff
@@ -0,0 +1,294 @@
...
+		let ratingsPromptEnabled: boolean = RatingsHelper.ratingsPromptEnabledForClient(clipperState.clientInfo.clipperType);
+		logEventInfo.ratingsPromptEnabledForClient = ratingsPromptEnabled;
+		if (!ratingsPromptEnabled) {
+			return false;
+		}
+
+		let doNotPromptRatingsStr: string = Clipper.Storage.getCachedValue(Constants.StorageKeys.doNotPromptRatings);
+		let lastBadRatingDateAsStr: string = Clipper.Storage.getCachedValue(Constants.StorageKeys.lastBadRatingDate);
+		let lastBadRatingVersion: string = Clipper.Storage.getCachedValue(Constants.StorageKeys.lastBadRatingVersion);
+		let lastSeenVersion: string = Clipper.Storage.getCachedValue(Constants.StorageKeys.lastSeenVersion);
```
VishaalK: This is gonna sound crappy but I am wary of this paradigm of knowing when a value is cached versus when it isn't. I think since you have tested it pretty thoroughly it will work here. The options are:
1. Always get the fresh value from `localStorage`, which means dealing with callbacks, which is why we switched to this design in the first place
2. Risk not having a fresh value. If the caller needs a fresh value, force them to use the async version.

What happens when you attempt `getCachedValue` but a corresponding `getValue` was never called? It is hard to differentiate between not being in storage, as we have a lot of cases where it doesn't exist yet, versus it not being in the cache. From `preCacheNeededValues` I know that these values are fine but we should discuss how we want to treat this in the future. 

TL;DR: `async` makes everything difficult -.-

VishaalK (approved):

iconix:
Noting a bug that @VishaalK informed me of: `height-animator` doesn't expand out enough anymore when on the `SignInNeeded` Panel (sign out of the clipper to see that the footer is cut off some - Feedback link is too close to the bottom)

#Pending -> #Resolved

src/scripts/clipperUI/ratingsHelper.ts
leakymemory: Time to take another hard look at being able to use async/await?

mttwc: ^ I think we should too. I think there's lots of benefit in this.

iconix: @VishaalK I definitely get your concern here - `getCachedValue` is easy to accidentally misuse.

@VishaalK @leakymemory @mttwc  I assume we're okay with using this paradigm for this PR, even if we need a better permanent solution... so how should we facilitate discussing the future?

Looks like `async/await` didn't make it into ES7 (ES2016), and only ["has a chance" of making ES8 (ES2017)](http://www.2ality.com/2016/02/ecmascript-2017.html)... so do we want to polyfill it? I can open a task for investigation if we all agree.

And what happened with the first hard look at `async/await`?

#Pending -> #Resolved (opened https://github.com/OneNoteDev/WebClipper/issues/43)

src/tests/clipperUI/ratingsHelper_tests.tsx
iconix: #Resolved (opened https://github.com/OneNoteDev/WebClipper/issues/41)

src/scripts/clipperUI/ratingsHelper.ts
iconix: #Resolved (in VSO)

src/tests/clipperUI/ratingsHelper_tests.tsx
iconix: 😃

src/scripts/clipperUI/panels/ratingsPanel.tsx
iconix: #WontFix after in-person discussion: any readability concern seems to apply only to the Init stage functions, and overall isn't too concerning

src/scripts/clipperUI/frontEndGlobals.ts
iconix: #Pending -> #Resolved (merged with master)

PR #23, opened 2016-09-13
Rename Region mode to 'Image' in Edge

### LINKED ISSUE #7: (Edge) Consider calling 'Region' mode 'Image' mode since the screenshot API is not supported yet
Edge does not support taking screenshots of the tab yet, so for clarity, we could call 'Region' mode 'Image' mode as it's more descriptive.

mttwc:
This is to be checked in when the matching string has been localized.

Fixes #7.


leakymemory (approved):
Looks good.

VishaalK (approved):
:sheep: it

mttwc:
Thanks everyone. I'm going to check this one in now because the string should be localized by the time we release the next version.

PR #24, opened 2016-09-14
If 10008 or 10013 is encountered on GetNotebooks, we should an error message...

### LINKED ISSUE #5: Provide a better message for O365 customers who have more than 20k items
We currently do not provide a good-enough message for O365 customers who are unable to save a new page as they reached the 20k item limit in OneDrive. See: https://github.com/OneNoteDev/WebClipper/blob/master/src/scripts/clipperUI/oneNoteApiUtils.ts

This is error code 20008, which we currently handle, but there error message is inadequate. Ideally we can provide a link to some page where the user can take actionable steps to reduce the number of items to < 20000.

mttwc:
… that contains a hyperlink to the blog post that tells the user how to remediate the issue. Currently uses an unscrubbed string that we're working on getting scrubbed. 

In a nutshell, the error string has curly braces around the word that will house the hyperlink, which the localization process will preserve. We then look for this word, whatever it is, and replace it with a hyperlinked version of itself.

Minor TODOs:
- Update comment where the error code is declared with the name once I find it
- Use scrubbed string when that is available to us
- Localize scrubbed string

This change relies on using onenotepicker 1.0.5 which has not been uploaded yet, see https://github.com/OneNoteDev/OneNotePicker-JS/pull/2.

Fixes #5


VishaalK (approved):
Lgtm.

A quick note: in tests, I notice we use a lot of utilities to build the expect result, such as `"Hi {OneNote}".replace("{OneNote}", expectedAnchor.outerHTML)`.

My initial feeling is that it makes it hard to see what the code is supposed to output, because you have to decipher two things. However, I can see that it makes it easier to write the test. 

For example, it could be changed to `Hi <a href="blah">OneNote</a>", which isn't too tedious to type and makes the test a bit clearer.

mttwc:
Good point, I'll make the change.

VishaalK:
^ Argh. I don't think I can edit my approval.

I wasn't asking for a change, just trying to see what your thoughts were (and others went they review this). 

Anyways, maybe I got my answer because Github interpreted my text as HTML, but I tried writing `Hi <a href="blah">OneNote</a>` :stuck_out_tongue_winking_eye:

mttwc:
My thoughts: I think in this case, readability is more important than making the test modifiable (we don't expect to modify the test anytime soon)

src/scripts/clipperUI/oneNoteApiUtils.ts
```diff
@@ -70,14 +118,16 @@ export module OneNoteApiUtils {
...
+	 */
 	function getResponseCodeInformation(apiResponseCode: string): { message: string, isRetryable: boolean, isExpected: boolean } {
 		let handledExtendedResponseCodes = {
 			"10001": { message: Localization.getLocalizedString("WebClipper.Error.GenericError"), isRetryable: true, isExpected: true }, // UnexpectedServerError
 			"10002": { message: Localization.getLocalizedString("WebClipper.Error.GenericError"), isRetryable: true, isExpected: true }, // ServiceUnavailable
 			"10004": { message: Localization.getLocalizedString("WebClipper.Error.PasswordProtected"), isRetryable: false, isExpected: true }, // PasswordProtectedSection
 			"10006": { message: Localization.getLocalizedString("WebClipper.Error.CorruptedSection"), isRetryable: false, isExpected: true }, // CorruptedSection
 			"10007": { message: Localization.getLocalizedString("WebClipper.Error.GenericError"), isRetryable: true, isExpected: true }, // ServerTooBusy
-			"10008": { message: Localization.getLocalizedString("WebClipper.SectionPicker.NotebookLoadUnretryableFailureMessage"), isRetryable: false, isExpected: true }, // SharepointQueryThrottled
```
iconix, line 80: I think removing this will cause `UnhandledApiCode` failures if `10008` is ever the error being handled and logged (like in the `GetNotebooks` event)

src/scripts/clipperUI/oneNoteApiUtils.ts
```diff
@@ -70,14 +118,16 @@ export module OneNoteApiUtils {
 		return responseCodeInfo ? responseCodeInfo.isRetryable : false;
 	}
 
+	/**
+	 * Retrieves response code information given that the context is in POSTing a clip.
```
iconix, line 101: Not sure this is strictly true - doesn't sound like `30104/UserAccountSuspended`, `30105/OneDriveForBusinessNotProvisioned`, and maybe some others, will necessarily occur when posting a clip

In general, note that we use `oneNoteApiUtils.ts` to handle errors for more than just the `ClipToOneNoteAction` event

src/scripts/clipperUI/components/sectionPicker.tsx
```diff
@@ -288,10 +288,9 @@ export class SectionPickerClass extends ComponentBase<SectionPickerState, Sectio
 			}
 		}
 
-		// Display an un-retryable message if API responds with a code that is not retryable
-		// Note that if this is an unhandled api response code, we assume it is not retryable
-		if (this.state.apiResponseCode && !OneNoteApiUtils.isRetryable(this.state.apiResponseCode)) {
-			localizedStrings.notebookLoadFailureMessage = Localization.getLocalizedString("WebClipper.SectionPicker.NotebookLoadUnretryableFailureMessage");
+		// If we can show a better message, especially an actionable one, we do
+		if (this.state.apiResponseCode && OneNoteApiUtils.getLocalizedErrorMessage(this.state.apiResponseCode)) {
```
iconix: why the removal of the `!isRetryable` check?

without it, it seems like if the API responds with a _retryable_ API response that isn't "actionable", then we're going to fallback to the `NotebookLoadUnretryableFailureMessage`, instead of the correct fallback of `NotebookLoadFailureMessage`. Might be missing something...

mttwc: I removed it because I figured that with the sectionPicker, there's no concept of 'retrying' a GetNotebooks call. Would it be a better experience if we show the retryable message for retryable codes anyway?

src/scripts/clipperUI/oneNoteApiUtils.ts
mttwc: Hmm, I think I'll do a minor cleanup of these methods. Previously, this method held messages for both the clip scenario and GetNotebooks scenario and I wanted to seperate them out so we don't show one message for the other scenario. Perhaps there's a better way of doing things.

src/scripts/clipperUI/oneNoteApiUtils.ts
mttwc: See below. I'll try and clean up this file.

mttwc:
I'll be working on adding the scrubbed string.

mttwc:
The string has been updated to the scrubbed version. Once again, since the link is aka.ms, we can change the url any time without having to change code. Looking to check in the string for localizers today. This should only be merged when localization is complete.

src/scripts/clipperUI/components/sectionPicker.tsx
iconix: Couldn't we get a `503/Service Unavailable` (API code `10002`) with the GetNotebooks call, for example? That would be retryable (in which I mean, it would help if the user would try the Clipper again later).

mttwc:
Updated this PR to use two strings instead of the one. Still need to work on @iconix's suggestions/comments.

mttwc:
Added some old functionality back as per @iconix's comments. Ready for review.

PR #29, opened 2016-09-16
Remove DOM event listener attributes from selections

### LINKED ISSUE #27: onload attribute not being removed from elements in a selection clip
This causes an error saying we refuse to execute the inline event handler. This shouldn't break us, but we should remove all attribute-based event handlers from selections.

mttwc:
Fixes #27. 'Benchmarked' against sites with a whole lot of elements (NYTimes, SeattleTimes) by selecting everything on the page. Seems to super quick (<10ms).

### DIFF
--- a/src/scripts/domParsers/domUtils.ts
+++ b/src/scripts/domParsers/domUtils.ts
@@ -728,6 +728,47 @@ export module DomUtils {
 		return a;
 	}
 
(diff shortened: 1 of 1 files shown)


leakymemory (approved):

VishaalK:
I too approve these changes

PR #32, opened 2016-09-16
Replace center elements with divs

### LINKED ISSUE #30: HTML center should be replaced with div for selections
ONML does not allow the user of <center> so we should replace it with a plain old div

mttwc:
ONML does not support `<center>` so we convert them to `<div>`. Fixes #30


leakymemory (approved):
Looks good.

VishaalK:
Looks great.

PR #33, opened 2016-09-16
[Feature] Polish the UI according to designers' critiques

VishaalK:
I collected the feedback we got in the last few e-mails and tracked them in VSO. Here are the tasks I completed:
- Change font size controls to two separate buttons instead of a toggle 
- Move Chevron in Section Picker title component slightly to the left ([corresponding PR in OneNotePicker-JS](https://github.com/OneNoteDev/OneNotePicker-JS/pull/3))
- Double check the font-weight on the location dropdown and the clip buttons on Mac OS X versus Windows (font looks lighter, when it shouldn't)
- Remove feedback link and account dropdown in Clip Successful screen  
- Change Serif Toggle fill to #FFFFFF and the font-color to #7321A6 
- Change region screenshot icon back to old one
- Change icon for Product to Aaron's new icon


mttwc:
Looks good. You should look into using FileOptimizer to compress the images as well :)

leakymemory (approved):
:+1:

PR #40, opened 2016-09-20
Prevent highlighting in non-editable modes

### LINKED ISSUE #38: Bookmark mode is highlightable
1) Turn on highlighter in article mode
2) Switch to bookmark mode
3) Select some text in the bookmark

Expected: Nothing happens
Actual: That line gets highlighted

mttwc:
This should fix the root cause of the present and past highlighting issues we've had.

Currently, we attach the highlighter to the previewBody element, which happens to be present across all preview modes. This is an issue because it's possible to get into weird states where you can highlight things in other modes, which again leads to even stranger behavior. For 'editable' previews, i.e., selection and augmentation, the highlighter will now be attached to the highlightablePreviewBody element, which is ONLY rendered in these modes.

i.e.,

``` html
<div id="previewBody ...>
  <div id="highlightablePreviewBody ...>
    {previewContent}
  </div>
</div>
```

As opposed to non editable modes:

``` html
<div id="previewBody ...>
  {previewContent}
</div>
```

Fixes #38.


leakymemory:
Nice change.

PR #42, opened 2016-09-20
Lint, lint, lint

larsendaniel:
Just cleaning up some ReSharper lint, because, well, why not.  :D


mttwc:
Looks good to me! 🐑 🐑 🐑

PR #47, opened 2016-09-22
Feature/typescript2.0

larsendaniel:
Upgrade to the TypeScript 2.0 compiler
Use the new 2.0 @types definitions (via npm), and remove the need for the `gulp setup` step and extra `typings` folder
Fix a few very minor compiler warnings with the new compiler


src/scripts/extensions/clipperInject.ts
```diff
@@ -228,7 +226,7 @@ export class ClipperInject extends FrameInjectBase<ClipperInjectOptions> {
 				this.uiCommunicator.callRemoteFunction(Constants.FunctionKeys.escHandler);
 			}
 			if (oldOnKeyDown) {
-				oldOnKeyDown(event);
+				oldOnKeyDown.call(document, event);
```
larsendaniel, line 229: The new compiler was complaining about this line, I _think_ I resolved it the correct way, but having someone double check would be nice ;)

leakymemory: I haven't found a ton of references to this yet, though I did find one that described doing it this way.  We'll run it through the gambit of browsers and make sure everything is gravy.  Thanks!

leakymemory:
These changes look really great - thanks for doing this Daniel!

leakymemory:
Testing note: I checked out this branch and things seem to be building and running just fine.  I think we're good to merge whenever the timing works out.

mttwc:
We've added a few more libraries recently that will need to be ported over to the new typings system.

larsendaniel:
I added `sanitize-html` with my latest merge.  Were there any others?

mttwc:
I think that's it! I was comparing tsconfig.json and the old typings.json and it looked like we had a few typings declarations that weren't being used :P

PR #53, opened 2016-09-26
Fixed region label vertical alignment

### LINKED ISSUE #44: (master) 'Region' mode label is not aligned with other mode labels

mttwc:
Fixes #44

### DIFF
--- a/src/styles/clipper.less
+++ b/src/styles/clipper.less
@@ -142,6 +142,13 @@
 	}
 }
 
+#regionButton {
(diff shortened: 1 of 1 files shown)


mttwc:
Note setting width to 24px will NOT work as it will cause the regionButton img to become its old self 'i.e., it will look smaller than the other icons'

VishaalK:
^ Good to know. Looks good. Thanks for the fix!!

PR #54, opened 2016-09-27
Add query param so OneNote Online knows it was navigated to from the Web Clipper

mttwc:


mttwc:
Closing this ... I think there's a better way, which I will push up.

mttwc:
Reopening this because we might find it useful one day

leakymemory:
Seems pretty straight forward.

PR #55, opened 2016-09-27
Added the ability for the extension to inject an install marker on frames

mttwc:
We need a way for our sites to reliably check if the Web Clipper is installed in a browser agnostic manner. On navigates, we inject an empty, hidden div onto the page and its frames which we can then check for.


mttwc:
Temporarily closing this at it turns out this doesn't execute as soon as the body is ready, which is what we really need.

PR #56, opened 2016-09-28
Append a div marking the Web Clipper is installed on this browser in all frames

mttwc:
As soon as the HTML element is made available, we append a div to the documentElement to mark that the user has the Web Clipper installed. This happens before head scripts are executed. Our sites like ONO or even the install page can now detect if the user has installed the Web Clipper and react accordingly.


leakymemory:
Great!  This is something that will help a bunch in the future.

mttwc:
Verified working in Chrome, Firefox, Edge, Safari

PR #58, opened 2016-09-28
[Bug] Wrap all `.innerHTML = ` calls with `DOMPurify.sanitize`

### LINKED ISSUE #46: Reduce and/or eliminate use of `.innerHTML = ` in the code
To get into the Mozilla Add-ons Store, we need to reduce our use of this as it is a potential security issue.

There are 5 places in the code where we do this:
1. `logPageModifications(...)` in `saveToOneNote.ts`: we create a DIV in the document and put  `clipperState.augmentationPreviewInfo.previewBodyHtml` in to determine how many highlights are in the text. It is pretty tedious to determine this information in memory, Also `previewBodyHtml` is a big ol' string that needs to be converted to HTML in order for this work.
   **Update**: Shouldn't matter if we sanitize HTML.  
2. `createPostProcessedHtml(...)` sticks `previewBodyHtml` into a `div` so we can apply the `font-size` and `font-family` in the preview as in-line styles. Same issue as above.
   **Update**: Shouldn't matter if we sanitize HTML.
3. `removeElementsNotSupportedInOnml(...)` in `domUtil.ts`. We replace the `<center>` tag with a div with the same `innerHTML`. Interesting StackOverflow: http://stackoverflow.com/questions/3435871/jquery-how-to-change-tag-name. A common theme is using an already existing node and traversing it using the DOM APIs. 
   **Update**: Shouldn't matter if we sanitize HTML, we can do it with `sanitize-html`. 
4. `getDocumentFromDomString(...)` in `domUtils.ts`. Used in the bookmark mode to get the content of the page.
   **Update**: We use this function retrieve `img`, `meta`, and `link` tags from the original, non-augmented `contentData` of the page. If possible, we could get it from the augmentation result, which is guaranteed to be cleaned. We are rendering the contentData into a document on our iframe, which could be potentially malicious. 
5. When we use `Selection` mode from the Context menu, we wrap whatever they've selected in a div.  We convert a `RangyRange` into a `string` by calling `.toHtml()`. Maybe we can walk the RangyRange instead?
   **Update**: this one is the most difficult one, but I think we can actually find a way to wrap this using pure JS. 

I've identified ways we can sanitize the HTML before injecting it anywhere, but it might not be necessary.

The "outstanding" issue is that we get `ContentInHtml` from our Augmentation API and then put that in the preview. That comes back as a string, so we have to parse that at some point.

VishaalK:
We use https://github.com/cure53/DOMPurify as suggested by Mozilla to sanitize all the raw html we are passing around. We use a whitelist (https://dev.onenote.com/docs#/introduction/html-tag-support-for-pages) of tags and attributes.

DOMPurify has a bug where it sometimes infinitely loops on tables. We have a repro on Amazon product pages, and a workaround in the code, but I have to get an isolated repro and submit that to DOMPurify. 

Fixes #46


mttwc:
Closing this as its currently broken on pages with tables.

PR #59, opened 2016-09-28
Fix blank image recognition race condition

### LINKED ISSUE #50: USA Today page Augmentation API call is successful, but Article preview shows endless spinner
http://www.usatoday.com/story/news/nation/2016/09/22/state-emergency-charlotte-protests-continue-second-night/90825016/

mttwc:
Fixes #50

### DIFF
--- a/src/scripts/domParsers/domUtils.ts
+++ b/src/scripts/domParsers/domUtils.ts
@@ -360,7 +360,6 @@ export module DomUtils {
 				// Just passing in the image node won't work as it won't render properly
 				// and the algorithm will think every pixel is (0,0,0,0)
 				let theImg = new Image();
-				theImg.src = img.src;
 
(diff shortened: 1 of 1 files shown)


src/scripts/domParsers/domUtils.ts
```diff
@@ -372,6 +371,8 @@ export module DomUtils {
 				theImg.onerror = () => {
 					resolve(undefined);
 				};
+
+				theImg.src = img.src;
```
leakymemory, line 376: Do you think that it's worth a comment here noting the importance of setting the src property _after_ everything else?

leakymemory:
Interesting.  Looks good.

src/scripts/domParsers/domUtils.ts
mttwc: Good idea. Will add it in.

PR #62, opened 2016-09-29
(Fix for xhtml sites) Enforce the use of lowercase when creating or querying for html elements

### LINKED ISSUE #61: Exception thrown when clipping http://us.battle.net/d3/en/profile/leshin-1101/hero/8664666
chromeInject.js:7137 Uncaught TypeError: Cannot read property 'postMessage' of undefined

mttwc:
Turns out the Clipper won't work on xhtml sites such as http://us.battle.net/d3/en/profile/leshin-1101/hero/8664666

This is because xhtml does not recognize uppercase elements, i.e., IFRAME won't work, but iframe does. Fixes #61


leakymemory:
The change looks good.  Not sure why we went with uppercase in the first place...

PR #66, opened 2016-09-29
[Bug] Wrap all `.innerHTML = ` calls with `DOMPurify.sanitize`

VishaalK:
The Firefox review team pointed out that we have some unsafe lines of code. Specifically, whenever we set the `.innerHTML =` of an element it is a potential attack vector. In order to mitigate this, they said we should "sanitize" the html we put in there. To sanitize it, we have a whitelist of allowed tags and attributes, and use the `sanitize-html` library to do this for us. Everything we put into  `.innerHTML =` lines is run through this function.

The `cleanHtml` function we wrote that utilizes the library lives in `DomUtil.ts`, along with the whitelist of tags and attributes.

Fixes #58


src/scripts/clipperUI/panels/successPanel.tsx
```diff
@@ -18,8 +18,7 @@ class SuccessPanelClass extends ComponentBase<{ }, ClipperStateProp> {
 		Clipper.logger.logUserFunnel(Log.Funnel.Label.ViewInWac);
 		let data = this.props.clipperState.oneNoteApiResult.data as OneNoteApi.Page;
 		if (data && data.links && data.links.oneNoteWebUrl && data.links.oneNoteWebUrl.href) {
-			let urlWithFromClipperParam = Utils.addUrlQueryValue(data.links.oneNoteWebUrl.href, Constants.Urls.QueryParams.wdFromClipper, "1");
-			window.open(urlWithFromClipperParam, "_blank");
+			window.open(data.links.oneNoteWebUrl.href, "_blank");
```
leakymemory: This is removing Matthew's change. Probably not what we want.

VishaalK: Indeed it is not what we want !

src/scripts/constants.ts
```diff
@@ -327,7 +327,6 @@ export module Constants {
 			export var timeoutInMs = "timeoutInMs";
 			export var url = "url";
 			export var userSessionId = "userSessionId";
-			export var wdFromClipper = "wdfromclipper"; // This naming convention is standard in OneNote Online
```
leakymemory, line 330: Should not be removed.

src/scripts/constants.ts
```diff
@@ -340,13 +340,13 @@ export module Constants {
 		export var maxClipSuccessForRatingsPrompt = 12;
 		export var maximumJSTimeValue = 1000 * 60 * 60 * 24 * 100000000; // 100M days in milliseconds, http://ecma-international.org/ecma-262/5.1/#sec-15.9.1.1
 		export var maximumFontSize = 72;
-		export var maximumNumberOfTimesToShowTooltips = 3;
+		export var maximumNumberOfTimesToShowTooltips = 3;
```
leakymemory: Nit.

src/scripts/constants.ts
```diff
@@ -340,13 +340,13 @@ export module Constants {
...
-		export var maximumNumberOfTimesToShowTooltips = 3;
+		export var maximumNumberOfTimesToShowTooltips = 3;		
 		export var minClipSuccessForRatingsPrompt = 4;
 		export var minimumFontSize = 8;
 		export var minTimeBetweenBadRatings = 1000 * 60 * 60 * 24 * 7 * 10; // 10 weeks
 		export var noOpTrackerTimeoutDuration = 20 * 1000; // 20 seconds
 		export var timeBetweenDifferentTooltips = 1000 * 60 * 60 * 24 * 7 * 1; // 1 week
-		export var timeBetweenSameTooltip = 1000 * 60 * 60 * 24 * 7 * 3; // 3 weeks
 		export var timeBetweenTooltips = 1000 * 60 * 60 * 24 * 7 * 3; // 21 days
+		export var timeBetweenSameTooltip = 1000 * 60 * 60 * 24 * 7 * 3; // 3 weeks
```
leakymemory: Nit. Keep alphabetical.

src/scripts/domParsers/domUtils.ts
```diff
@@ -13,34 +18,126 @@ import {VideoExtractorFactory} from "./VideoExtractorFactory";
...
+		export const table = "table";
+		export const td = "td";
+		export const title = "title";
+		export const tr = "tr";
+		export const u = "u";
+		export const ul = "ul";
 		export const video = "video";
 	}
 
+	const attributesAllowedByOnml: { [index: string]: string[] } = {
```
leakymemory, line 85: Could we link to the OneNote Dev Center page that lists the supported tags?

VishaalK: Sure!

src/scripts/domParsers/domUtils.ts
```diff
@@ -63,14 +160,42 @@ export module DomUtils {
 		Tags.video
 	];
 
-	export function removeElementsNotSupportedInOnml(doc: Document) {
+	/**
+	 * Given an HTML Document in string form, return an HTML Document in string form
+	 * with the attributes and the content between the HTML tags scrubbed, while preserving
+	 * document structure
+	 */
+	export function cleanHtml(contentInHtml: string): string {
```
leakymemory, line 170: Can we call this sanitizeHtml()?  I think the name would describe the functionality a little better.

Edit: Never mind. Namespace conflicts.

leakymemory:
Ok, looks good.

PR #67, opened 2016-09-29
Add in the default settings for the ratings prompt

leakymemory:
This fixes the unit test failures.


mttwc:
Merging this in. We'll discuss a more permanent solution to this when we can.

PR #69, opened 2016-09-30
Increase FullPageScreenshot timeout

### LINKED ISSUE #68: FullPageScreenshotCall has a tendency to time out on larger sites with lots of images
Example: http://html5doctor.com/the-figure-figcaption-elements/
For me, this is hitting the 30s timeout consistently.

mttwc:
We're hitting the full page screenshot 30s timeout consistently on certain sites, such as http://html5doctor.com/the-figure-figcaption-elements/
This would be a combination of things like upload/download time, and processing time on the server.

I think we should up our timeout so users don't get frustrated on certain sites. Would 50s be enough?

Fixes #68

### DIFF
--- a/src/scripts/contentCapture/fullPageScreenshotHelper.ts
+++ b/src/scripts/contentCapture/fullPageScreenshotHelper.ts
@@ -23,6 +23,8 @@ export interface FullPageScreenshotResult extends CaptureFailureInfo {
 }
 
 export class FullPageScreenshotHelper {
+	private static timeout = 50000;
+
 	public static getFullPageScreenshot(pageInfoContentData: string): Promise<FullPageScreenshotResult> {
 		return new Promise<FullPageScreenshotResult>((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


iconix:
So, this is just a call for the preview image... are users going to be willing to wait longer than 30s for a preview to show up anyway? I would think that most likely, users will get impatient and clip without waiting for the full timeout - whether it be 30s or 50s

Btw, mainly pushing back because we're special-casing this call from the standard 30s timeout :) Want to make sure we have a good reason to do so

leakymemory:
Even waiting 30 seconds is pretty painful - though I think I'll take even more painful over broken. :-)  Let's go with the 50 second bump right now and see how that goes.  I'd like to dig in a little and do some bandwidth throttling in test to see what the experience is more realistically like.  This is a somewhat unique call in that it results in a potentially large download.  The PDF upload might be another place to take a look.

As to whether we need to bump the standard timeout?  We should look at the numbers and see how often the timeout is getting hit.  I don't want to lull ourselves into thinking that all is well in the world just because things are working (albeit taking a really long time).

PR #71, opened 2016-09-30
[Feature] Add a "nominify" option so we can build a production version without minifying 

VishaalK:
This is for the initial submission to Firefox. It might prove useful in the future too


mttwc:
As discussed, let's keep production and nominify separate :)

VishaalK:
^ As discussed again lol, I'll leave it the way it is and wait until we have more build rules and a need arises.

iconix (approved):
lgtm. personally, I prefer the `nominify` option

PR #72, opened 2016-09-30
[Bug] Ratings prompt UT fix

iconix:
1. Fix ratings panel UTs that fail when settings are not included
2. Revert [workaround](https://github.com/OneNoteDev/WebClipper/pull/67) for failing UTs
3. Other minor test cleanup
## 

If curious: basically, the problem was that, depending on the way we build, the failing tests could either begin with empty prod settings or non-empty prod settings.

There were two problems: 1) using **prod** (instead of mock) settings in the tests, and 2) including an **extra (and nonstandard) panel transition** from `Init` to `Rate/Feedback`.

Broken behavior flow:

> empty _prod_ settings => test transitions from `Init` to `End`
> non-empty _prod_ settings => test transitions from `Init` to `Rate/Feedback` to `None`

Fixed behavior flow:

> empty _mock_ settings => test transitions from `Rate/Feedback` to `None`

The empty mocks are enforced with the `Settings.setSettingsJsonForTesting({})` call in `beforeEach`. Starting at the `Rate/Feedback` panel (instead of `Init`) is setup with the `controllerInstance.setState({...})` calls.

### DIFF
--- a/src/settings/default.json
+++ b/src/settings/default.json
@@ -63,7 +63,7 @@
 			"^(http(s)?)://(paleoleap.com)/([^/]*-[^/]*)/$",
 			"^(http(s)?)://(pinchofyum.com)/([^/]*-[^/]*)$",
 			"^(http(s)?)://(recipe\\.rakuten\\.co\\.jp)()/recipe/[0-9]+/$",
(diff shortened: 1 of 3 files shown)


leakymemory:
Looks good.  Thanks for the follow-up on this.

mttwc:
Looks good to me! :shipit:

Issue #77, opened 2016-10-04
Long citations get truncated on overflow and become hard to read and copy


iconix:
A small user-friendliness task.

e.g., the "Clipped from" urls here:

![image](https://cloud.githubusercontent.com/assets/1144401/19089684/43c3c16c-8a30-11e6-9a75-4e9bf6898c23.png)

![image](https://cloud.githubusercontent.com/assets/1144401/19089733/6e19724a-8a30-11e6-81c7-5182ad4c9114.png)

Hard for the user to interact with the full URL, if desired.

PR #78, opened 2016-10-04
[Bug] Make long citations copy-able/fully readable

iconix:
1. On hover, the full URL will show up (thanks to the `title` property)
2. If a user wants to copy the URL, they can right click -> copy link address

Fixes https://github.com/OneNoteDev/WebClipper/issues/77

### DIFF
--- a/src/scripts/clipperUI/components/previewViewer/previewComponentBase.tsx
+++ b/src/scripts/clipperUI/components/previewViewer/previewComponentBase.tsx
@@ -83,7 +83,7 @@ export abstract class PreviewComponentBase<TState, TProps extends ClipperStatePr
 				{this.props.clipperState.currentMode.get() !== ClipMode.Bookmark
 					? <div id={Constants.Ids.previewUrlContainer}>
 						{sourceUrlCitationPrefix}
-						<a href={sourceUrl} target="_blank">{sourceUrl}</a>
+						<a href={sourceUrl} target="_blank" title={sourceUrl}>{sourceUrl}</a>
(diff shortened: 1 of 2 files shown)


src/scripts/clipperUI/components/previewViewer/previewComponentBase.tsx
```diff
@@ -70,8 +70,10 @@ export abstract class PreviewComponentBase<TState, TProps extends ClipperStatePr
 	}
 
 	private getPreviewSubtitle(): any {
-		let sourceUrlCitation = Localization.getLocalizedString("WebClipper.FromCitation")
-			.replace("{0}", this.props.clipperState.pageInfo ? this.props.clipperState.pageInfo.rawUrl : "");
+		let sourceUrlCitationPrefix = Localization.getLocalizedString("WebClipper.FromCitation")
+			.replace("{0}", ""); // TODO can we change this loc string to remove the {0}?
```
iconix, line 74: I was concerned about just making this change to remove the `{0}` from `WebClipper.FromCitation` (on both the client and server) - are there any weird timing issues to be aware of from a localization standpoint?

mttwc: Looks like in Javascript, if the 1st param of replace does not exist in the string, nothing happens. I think this change is fine, and we should have no problems changing the one stored on the server without having to worry about n-1 issues.

mttwc (approved):
Looks good. Thanks for doing this!

PR #79, opened 2016-10-05
Update all of the packages to the latest versions

leakymemory:
I went through and updated all of the packages to their most recent
versions and then basically "hardened" them on that version.  We're having
a problem where different machines are generating different code based on
when the last time they ran npm install was.  This has been causing
problems during our Mozilla code review, and is generally scary as we each
build and could potentially deploy.  This way we are all guaranteed to all
be on the same version.

This also means we need to be more diligent in making sure that the
package stay up-to-date going forward. There are probably good ways of
doing this so I'll dig a little more into what this really means.

*Note that the QUnit related packages are not updated. This is because
there are some recent breaking changes that will need to be dealt with
separately.


VishaalK:
I checked out this branch locally to test it. Everything built correctly after "npm install" and I verified behavior by using the Clipper on Medium, YouTube, Amazon, Vimeo, etc. and it worked. Ship it!

Update: It also works on Firefox.

PR #82, opened 2016-10-06
[Feature] Update the Aria lib and make sure --nominify applies to this lib. 

VishaalK:
According to the release notes, there is only a minor bug fix with timezones, which we don't rely on. 

I verified on Chrome and Firefox.

Also you need to pull down the correct branch of our internal repository for this to work. <-- wondering if this is a smell.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -59,6 +59,8 @@ var PATHS = {
 
 var RTL_SUFFIX = "-rtl";
 
+var ARIA_LIB_VERSION = "2.7.1";
+
 // Used for debugging glob declarations
 function printGlobResults(glob) {
 	globby.sync(glob).map(function(filePath) {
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -433,8 +433,9 @@ function exportCommonJS(targetDir) {
 
 		var logManagerExportTask;
 		if (fileExists(PATHS.BUNDLEROOT + "logManager_internal.js") && !argv.nointernal) {
+			let ariaLibToInclude = argv.nominify ? "aria-web-telemetry-2.7.1" : "aria-web-telemetry-2.7.1.min.js";
```
leakymemory: What would you think about pulling the version number out into a variable and then using that to build the file name?  It might make things a little cleaner going forward (updating, etc.).

leakymemory (approved):
Looks good.

gulpfile.js
VishaalK: That's a great idea! I'll do this now before I merge

PR #84, opened 2016-10-06
Fix the parsing of the error description.

### LINKED ISSUE #83: O365 Sign-in failures are not logging the error description
It looks like we're not correctly logging the error description when O365 sign-in errors occur.

leakymemory:
It looks like O365 returns a slightly different query parameter for the error description field.  This is a slight modification to the constant regular expression such that it matches both versions.

Fixes #83


iconix (approved):
📈

mttwc:
Nice fix!

PR #88, opened 2016-10-07
Fix bad logic in videoDomainIfSupported and whitelisteddomains

### LINKED ISSUE #86: [Bug] Navigating to our internal logging page shows the Video tooltip

mttwc:
One of the checks in videoDomainIfSupported is incorrectly returning true if the domain name ended in 0, 1 or 2. This is because we iterated over the enums (which are numbers) and treated them as strings, thinking it was the string version of the enum. Added UTs to catch this case. 

Also, found a bug where we were incorrectly thinking that a domain is supported if the ending of the domain was that substring. We should be checking that the entire domain name is supported, not just the end substring, else domain names like "123vimeo.com" will pass our tests.

Fixes #86. Also, happy 2000 assertions!


src/tests/domParsers/videoUtils_tests.ts
```diff
@@ -1,73 +1,119 @@
...
+		"https://vimeo.com/album/45196609/page:1",
+		"https://vimeo.com/album/45196609/page:3/sort:preset/format:thumbnail",
+		"https://vimeo.com/album/45196609/sort:preset/format:thumbnail/page:2",
+		"https://vimeo.com/album/45196609/video/45196609",
+		"https://vimeo.com/ondemand/45196609"
+	];
+	for (let pageUrl of supportedVimeoUrls) {
 		let domain = VideoUtils.videoDomainIfSupported(pageUrl);
-		strictEqual(domain, undefined, pageUrl + " should NOT be supported");
+		strictEqual(domain,	"Vimeo", pageUrl + " should return Vimeo");
```
VishaalK: nit: weird spacing

src/tests/domParsers/videoUtils_tests.ts
```diff
@@ -1,73 +1,119 @@
...
+		"https://www.youtube.com/watch?v=dQw4w9WgXcQ#foo",
+		"https://www.youtube.com/watch?feature=youtu.be&t=30s&v=dQw4w9WgXcQ",
+		"https://m.youtube.com/watch?v=dQw4w9WgXcQ&feature=youtu.be",
+		"https://www.youtube.com/embed/dQw4w9WgXcQ"
+	];
+
+	for (let pageUrl of supportedYouTubeUrls) {
 		let domain = VideoUtils.videoDomainIfSupported(pageUrl);
-		strictEqual(domain, undefined, pageUrl + " should NOT be supported");
+		strictEqual(domain,	"YouTube", pageUrl + " should return YouTube");
```
VishaalK: nit: same weird spacing

VishaalK:
Sheep it

PR #92, opened 2016-10-11
[Bug] Update ratings prompt logic for delayed client switch ons

iconix:
Ensuring ratings prompt shows as intended for users on initially disabled clients (e.g., Firefox).


mttwc:
Thanks for doing this! Looking into this now. Just a reminder to add a corresponding Issue in the issue tracker so in the future we know that this has been tackled :)

leakymemory (approved):
Yep, looks good!

PR #94, opened 2016-10-12
Log the server correlation id for /userinfo

leakymemory:
I feel like there needs to be a deeper overhaul of how we log our XHR
requests, but this takes care of the immediate need of being able to trace
the /userinfo calls on the server.  This is especially important in
tracking down the sign-in issues that are still hit on occasion.


src/scripts/extensions/authenticationHelper.ts
```diff
@@ -136,12 +136,18 @@ export class AuthenticationHelper {
 	/**
 	 * Makes a call to the authentication proxy to retrieve the user's information.
 	 */
-	public static retrieveUserInformation(clipperId: string, cookie: string = undefined): Promise<ResponsePackage<string>> {
+	public static retrieveUserInformation(clipperId: string, cookie: string = undefined, logger: Logger = undefined): Promise<ResponsePackage<string>> {
```
iconix: You could make this non-static so you have access to the logger without passing it in

src/scripts/extensions/authenticationHelper.ts
```diff
@@ -136,12 +136,18 @@ export class AuthenticationHelper {
...
 		return new Promise<ResponsePackage<string>>((resolve, reject: (error: OneNoteApi.RequestError) => void) => {
 			let userInfoUrl = Utils.addUrlQueryValue(Constants.Urls.Authentication.userInformationUrl, Constants.Urls.QueryParams.clipperId, clipperId);
 
+			let retrieveUserInformationEvent = new Log.Event.PromiseEvent(Log.Event.Label.RetrieveUserInformation);
```
iconix, line 143: Since you most actively look at this data, you would know best: might it be easier to debug auth issues if you added this new info to the existing `GetExistingUserInformation` event? Or is it helpful to keep them separate? There just seems to be large overlap between when these two events will be logged.

src/scripts/extensions/authenticationHelper.ts
```diff
@@ -154,7 +160,13 @@ export class AuthenticationHelper {
 				// The false case is expected behavior if the user has not signed in or credentials have expired
 				resolve({ parsedResponse: this.isValidUserInformationJsonString(response) ? response : undefined, request: request });
 			}, (error: OneNoteApi.RequestError) => {
+				retrieveUserInformationEvent.setStatus(Log.Status.Failed);
```
iconix, line 163: I know our naming convention `OneNoteApi.RequestError` is super confusing and overused (this request has nothing to do with the API), but still, you might want to replace these lines with a call to `OneNoteApiUtils.logOneNoteApiRequestError(retrieveUserInformationEvent, error);` for consistency

iconix (approved):
A few small suggestions, but otherwise lgtm!

src/scripts/extensions/authenticationHelper.ts
leakymemory: I like the idea of constancy, but in this case the OneNoteApiUtils is (currently) only available in UI thread.  I'm hesitant to pull this into the extension code unless there is good reason (and maybe even a little bit of refactoring).

src/scripts/extensions/authenticationHelper.ts
mttwc: Agreed. I think we should try to cut down the use of static methods where we can.

src/scripts/extensions/authenticationHelper.ts
mttwc: Added https://github.com/OneNoteDev/WebClipper/issues/98 to track the use of OneNoteApi.RequestError in places where it doesn't make sense.

mttwc:
I echo @iconix's comments. Looks good :)

PR #95, opened 2016-10-12
Allow selections to be clipped on PDFs

### LINKED ISSUE #52: Context menus does not work on PDFs
The webextensions API does not support this explicitly but we can potentially use a hack. This will need some investigation.

mttwc:
Added allowed documenturlpatterns to the Web Extensions context menu creator to allow for the different PDF plugins/extensions.

Allow the clip options pipeline to pipe selected text from the extension side.

This has been tested to work in Chrome and Firefox. I don't have Edge on me at the moment so I'll test it there when I can. This will not be supported in Safari as their API does not support the ability to grab selections outside of our Rangy workaround.

Fixes #52


leakymemory (approved):

mttwc:
Please note that Edge is bugged/has not implemented selections fully so it won't be able to take advantage of this.

PR #97, opened 2016-10-12
Fix Edge PDF view & clip

leakymemory:
It looks like in Edge, the embed element has an undefined type for PDFs.
We're going to evaluate the value of this check in the first place, but
fixing the logic such that the file extension (PDF) returns the
EnhancedUrl type is the right thing to do.


mttwc:
Looks good!

PR #101, opened 2016-10-13
Added comments to use of any-casting, or refactored it to not use it

### LINKED ISSUE #100: Casting to <any> should be accompanied with a comment
Yesterday @leakymemory and I found a bug that was a result of casting to <any>, and that code did not have a comment as to why that cast was needed. I think we should look through our code for these occurances and comment these (assuming casting to `<any>` is something we really need).

Of course, casting to `<any>` is pretty dangerous in the sense that it goes against what Typescript is for, and where possible, we should create interfaces and use those instead.

mttwc:
Fixes #100


leakymemory (approved):
This is a good change.  It would also be interesting if we could find a way to turn on the  noImplicitAny flag and lock down things even more.

Issue #103, opened 2016-10-14
Calls to Clipper.getUserSessionId() occasionally are not returning the correct value


leakymemory:
We have seen cases where the first two augmentation calls are sending up a session ID that doesn't correlate with the one being sent to our logging endpoint.

mttwc:
Clipper.getUserSessionId() is returning the right thing. We were setting the session id in the request as correlation id. I have a PR out that properly sets the correlation id and session id on early requests (Augmentation, FullPage) with their corresponding values.

PR #109, opened 2016-10-14
On Augmentation and FullPage API calls, we correctly send correlation id and session id as headers

### LINKED ISSUE #106: Ensure that correlation id is being logged in all requests to our endpoints

mttwc:
We had to modify some tests for this. The reason why server.respond() was failing was because it assumes that you already sent the XHR. With this change, sending the XHR happens after an async operation, so we end up with a race condition. This is fixed with using server.respondImmediately = true instead. However, I would expect that using that instead of server.respond() would work for all our other tests that involve fake servers, but it causes them to fail if they switch. Some further investigation is necessary on what 'the right way' is, but in the meantime, this will give us the extra logging we need.

Fixes #106 and #103


src/scripts/clipperUI/frontEndGlobals.ts
```diff
@@ -21,6 +21,19 @@ export class Clipper {
...
 
+	public static getUserSessionIdWhenDefined(): Promise<string> {
+		return new Promise<string>((resolve) => {
+			let sessionId = Clipper.sessionId.get();
+			if (sessionId) {
+				resolve(sessionId);
+			} else {
+				Clipper.sessionId.subscribe((definedSessionId: string) => {
+					resolve(definedSessionId);
+				}, { times: 1, callOnSubscribe: false });
```
leakymemory, line 32: Such an elegant solution!

leakymemory (approved):

PR #111, opened 2016-10-14
Long pdf clip message

mttwc:
Putting this PR out there, _though this has not been styled to redlines_.

The Clipping Pdf progress panel now shows a message after 5s of delay. Added sinon faketimers tests.


src/scripts/clipperUI/mainController.tsx
```diff
@@ -256,6 +257,10 @@ export class MainControllerClass extends ComponentBase<MainControllerState, Main
 			case PanelType.RegionInstructions:
 				return <RegionSelectingPanel clipperState={this.props.clipperState} />;
 			case PanelType.ClippingToApi:
+				if (this.props.clipperState.currentMode.get() === ClipMode.Pdf) {
+					return <ClippingPanelWithDelayedMessage clipperState={this.props.clipperState}
```
leakymemory, line 261: We should do this as a separate change, but I'd like to add a more generic version of this for any of the non-PDF clips that take more than a few seconds to come back.  Filed #116 to track.

leakymemory (approved):
Nice. This is something I've wanted for a while now.

src/scripts/clipperUI/mainController.tsx
mttwc: Good idea!

VishaalK (approved):
I also learned how to use clocks today. Thanks Matthew !

PR #115, opened 2016-10-14
Replace server.respond() calls with server.respondImmediately = true

### LINKED ISSUE #110: Replace server.respond() with something that can deal with asynchronicity
Calling server.respond() assumes the XHR has already been sent. Sometimes this is not the case. We need something that can deal with async. server.respondImmediately = true does not seem to work in synchronous scenarios for some reason. I'd love to standardize the way we do fake-server tests.

mttwc:
`server.respond()` assumes that you have already sent a request, which can make async code blocks hard to test. Here's a scenario:

```
function fnBeingTested() {
    doSomethingInBackground().then(() => {
        console.log("foo");
        // send xhr
    });
}
```

Imagine you are testing the above function. You want to get your mock server to send a response, so you set up your response in your test and do the following:

```
fnBeingTested();
server.respond();
```

This is the expected order of execution:

```
"foo"
response
```

This is what happens:

```
response (no-op, because we haven't sent xhr yet)
"foo"
```

This is because doSomethingInBackground is an async function, which means as soon as it gets hit, we call the server.respond() before the Thenable. We now have a race condition, and `server.respond()` no longer cuts the cheese. Rather than doing this, we now use `server.respondImmediately = true`, which in a nutshell, sets up a listener for the XHR being sent before automatically sending the response.

This will also make our tests more readable as it will follow a very natural flow that is similar to how we actually write our code in our codebase. For examples, simply look at some of the before vs after tests in sectionPicker_tests.tsx :)

From now on, let's disallow the use of `server.respond()`.
Fixes #110

### DIFF
--- a/src/tests/clipperUI/components/sectionPicker_tests.tsx
+++ b/src/tests/clipperUI/components/sectionPicker_tests.tsx
@@ -200,6 +200,7 @@ QUnit.module("sectionPicker-sinon", {
 		};
 
 		server = sinon.fakeServer.create();
+		server.respondImmediately = true;
 	},
 	afterEach: () => {
 		xhr.restore();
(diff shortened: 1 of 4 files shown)


leakymemory (approved):
Thanks for digging into this!

PR #117, opened 2016-10-17
PDF mode scroll listener

mttwc:
When the user scrolls in pdf mode, we 'pop-in' the page numbers. If the user hasn't scrolled in 1 second, we fade out the page numbers over 0.5 seconds.

### DIFF
--- a/src/scripts/clipperUI/components/previewViewer/pdfPreview.tsx
+++ b/src/scripts/clipperUI/components/previewViewer/pdfPreview.tsx
@@ -22,18 +22,35 @@ import { PreviewViewerPdfHeader } from "./previewViewerPdfHeader";
 import * as _ from "lodash";
 
 interface PdfPreviewState {
-	fadeIn: boolean;
+	showPageNumbers: boolean;
 }
 
 class PdfPreview extends PreviewComponentBase<PdfPreviewState, ClipperStateProp> {
-	private static scrollListenerAdded: boolean = false; // done on purpose
+	private static scrollListenerAdded: boolean = false;
+	private static scrollListenerTimeout: number;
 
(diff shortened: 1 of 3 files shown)


src/scripts/clipperUI/components/previewViewer/pdfPreview.tsx
```diff
@@ -22,18 +22,35 @@ import { PreviewViewerPdfHeader } from "./previewViewerPdfHeader";
...
 
 	private addScrollListener() {
 		if (!PdfPreview.scrollListenerAdded) {
 			let previewContentContainer = document.getElementById("previewContentContainer");
 			if (!!previewContentContainer) {
+				// When we detect a scroll, show page numbers immediately.
+				// When the user doesn't scroll for some period of time, fade them out.
 				previewContentContainer.addEventListener("scroll", (ev) => {
-					console.log("scroll handler");
+					if (Utils.isNumeric(PdfPreview.scrollListenerTimeout)) {
```
VishaalK, line 39: You might be able to use lodash's debounce function so you. https://lodash.com/docs/4.16.4#debounce I think it does something similar to what we are trying to do here. 

Otherwise looks good! Thanks Matthew!! I learnt a little bit about how to use css animatinos today

VishaalK (approved):

src/scripts/clipperUI/components/previewViewer/pdfPreview.tsx
mttwc: Hmm I looked through what debounce does, and I don't think it's suitable here, at least if we try using it, it will look less readable (and possibly less performant).

PR #119, opened 2016-10-18
Persist page range in clipperstate

mttwc:
When the user switches modes, their page selection now persists.


VishaalK (approved):
Looks dope

PR #120, opened 2016-10-18
Fix scroll listener on mode switch

mttwc:
Add the scroll listener to the window and check that we have the correct target there instead of doing it on the element itself which will get re-rendered at some point. This will ensure we don't lose the scroll listener. Additionally, reworked when we attach the listener to ensure our reference to the state object is always the correct one and not an old one.


VishaalK (approved):
Coolio

PR #121, opened 2016-10-18
Fix tabbing for pdf header

mttwc:
Fixed a bug where the user is unable to input their page range selection using tabbing and keyboard input. Also added UT for tab order.


leakymemory (approved):

VishaalK:
Thanks for fixing this!

PR #123, opened 2016-10-19
Fix onenote branding in tooltip in rtl

### LINKED ISSUE #122: RTL: In popup, the Web Clipper icon and logo incorrectly appears on the top-left instead of the top-right
This causes the icon and logo to overlay the 'X'.

mttwc:
Fixes #122 

This was caused by:
- The image having a large amount of whitespace on the right, making it hard to style for RTL properly
- The css doesn't work in RTL


VishaalK:
Matthew Chiam holding it down for all RTL readers out there

PR #125, opened 2016-10-19
Add logging to pdf mode

mttwc:
Log the following events:
- On clip:
  - All page option vs page selection option
  - How many pages were specified in the selection box?
  - How many pages did the pdf contain?
  - Did the user attach the pdf?
  - Was it a local file?
- Log event if user encountered the panel asking them to turn on local file support that session

Also deleted unused code/comments.

While working on this, I created an issue for the cleanup of SaveToOneNote.ts as it's getting pretty big. It's low prio for now, but I think it's fairly high-prio after we finish up here.


VishaalK:
Awesome!

PR #128, opened 2016-10-20
Bucket PDF pages to PATCHes in such a way that they are as even as possible. Add a delay between PATCH.

mttwc:
Still needs to be cleaned up.


mttwc:
Cleaned up and finalized the timeout settings.

PR #133, opened 2016-10-21
Always enable add-a-note

mttwc:
I think we should do this across all our modes. It's a global state that exists across all modes, so we shouldn't gate this by a disable clip button.


leakymemory:
I honestly don't remember why we disabled it in the first place.  I guess the idea was that if you couldn't clip, then why let you add a note.

PR #134, opened 2016-10-21
PDF Mode Safari Styles Fix

VishaalK:


mttwc (approved):

mttwc:
I don't see this causing any issues in RTL, but have you tested this on other browsers?

src/styles/previewStyles.less
```diff
@@ -256,6 +256,7 @@
 					color: rgba(255,255,255,0.70);
 					line-height: 20px;
 					outline: none;
+					-webkit-user-select: auto;
```
leakymemory, line 256: Rather than using the webkit version, does it work if you just use the standard, "user-select" style?

src/styles/previewStyles.less
```diff
@@ -109,6 +110,12 @@
 					}
 				}
 
+				#pageRangeControl {
+					// Temporary workaround that fixes styling issues with the PDF Preview Header
```
leakymemory, line 111: Temporary?  Famous last words.  :-)

Is there something tracking the "real" fix here?  Also, how arbitrary is 300px?  Is that particular number based on anything?

leakymemory (approved):

src/styles/previewStyles.less
VishaalK: will attempt

src/styles/previewStyles.less
VishaalK: The 300px is large enough so that the width of the input as well as the radio buttons + labels all fit in the same box. Matthew brought up the "real" fix, which is some underlying differences in how Safari and Chrome handle CSS. I will create an issue for the real fix. I'd like to get this out the door

src/styles/previewStyles.less
VishaalK: Confirmed to work! Will delete those lines and push again.

PR #143, opened 2016-10-27
Add ProcessPdfIntoDataUrls event to new PDF mode

VishaalK:


src/scripts/clipperUI/saveToOneNote.ts
```diff
@@ -402,7 +402,14 @@ export class SaveToOneNote {
 	 */
 	private static createOneNotePagePatchRequest(pageId: string, pageIndices: number[]): Promise<any> {
 		let pdf = this.clipperState.pdfResult.data.get().pdf;
+		let getPageListAsDataUrlsEvent = new Log.Event.PromiseEvent(Log.Event.Label.ProcessPdfIntoDataUrls);
+		const numPages = pageIndices.length;
```
mttwc: There's some variables here that you declare that aren't being used anywhere else. We can just keep this concise and not do this.

mttwc:
🚢

leakymemory (approved):

src/scripts/clipperUI/saveToOneNote.ts
VishaalK: Will do!

PR #148, opened 2016-10-28
Remove unecessary subsequent PATCH permission checks

mttwc:
Currently, we call getPages() to check for PATCH permissions on every pdf clip, however, we know that if a user has already successfully checked once, or has signed in at some point in this version, that they have the latest scopes, so we store a value in localStorage to mark the device so. Of course, this is still not ideal, so I've added a lengthy comment, and I'll create an issue to track this so we can come up with a more permanent, reliable, and performant solution.


leakymemory (approved):
Ship it! :shipit:

Issue #144, opened 2016-10-27
Clean up Promise logic to use established patterns


VishaalK:
We use a lot of `new Promise<>((resolve, reject) => {})` when we can get rid of some of those in favor of `Promise.resolve(...)` and `Promise.reject(...)`.

VishaalK:
We can start in `saveToOneNote.ts`.

PR #152, opened 2016-10-28
Convert PATCH wait to Promise.all. Added a PromiseUtils module

mttwc:
Also got rid of url logging in the pdf screenshot helper.

### DIFF
--- a/src/scripts/clipperUI/saveToOneNote.ts
+++ b/src/scripts/clipperUI/saveToOneNote.ts
@@ -2,6 +2,7 @@
 
 import {ArrayUtils} from "../arrayUtils";
 import {Constants} from "../constants";
+import {PromiseUtils} from "../promiseUtils";
 import {Settings} from "../settings";
 import {StringUtils} from "../stringUtils";
 import {Utils} from "../utils";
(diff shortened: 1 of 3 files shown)


src/scripts/promiseUtils.ts
```diff
@@ -0,0 +1,12 @@
+export module PromiseUtils {
+	/**
+	 * Returns a promise that simply resolves after the specified time period
+	 */
+	export function wait(millieseconds: number): Promise<{}> {
+		return new Promise((resolve) => {
```
VishaalK, line 6: super cool!

src/scripts/clipperUI/saveToOneNote.ts
```diff
@@ -424,34 +425,24 @@ export class SaveToOneNote {
 		let timeBetweenPatchRequests = SaveToOneNote.timeBeforeFirstPatch;
 		return Promise.all([
 			indexesToBePatchedRanges.reduce((chainedPromise, currentIndexesRange) => {
-				return chainedPromise = chainedPromise.then((returnValueOfPreviousPromise /* should be a onenote response */) => {
+				return chainedPromise = chainedPromise.then(() => {
 					return new Promise((resolve, reject) => {
```
VishaalK, line 429: Since the `setTimeout` was taken out of this function, I think we can take out the Promise constructor and use `return Promise.resolve();` and `return Promise.reject(error);` instead. We'd have to use `return Promise.all([...`  and `return SaveToOneNote.createOneNotePageRequest(...` from within that. However, if it works right now, I think we should leave it and investigate that as part of #144. 

It might not even be necessary to use `return Promise.resolve()`, because we'll "propagate" the return value of `SaveToOneNote.createOneNotePagePatchRequest(...`.

VishaalK:
🚀

PR #170, opened 2016-10-31
Test pdf components

### LINKED ISSUE #149: Write component tests for new PDF components
There's a couple of small PDF components that we wrote for the new PDF mode. We need to write tests for them. Marking this as a bug because until we do so, we are not 100% confident of their correctness :)

mttwc:
Fixes #149


leakymemory (approved):
Looks good!

VishaalK:
spooky good 🎃

PR #180, opened 2016-10-31
Change error message on FullPageScreenshot failure, gray out clip button

### LINKED ISSUE #169: When content is not available, we should not inform users that "A preview isn't available, but you can still clip your page."
I think they got into the `"A preview isn't available, but you can still clip your page."` state with the `FullPageScreenshotContentFound` property set to `false` -- but it was false because there really was no content to clip.

So we should change messaging here to not mislead people into thinking they will still get a screenshot in their notebook.

mttwc:
If we are happy with this string, I can kick off localization.

Also I realized there was no UTs for optionsPanel. I have gone ahead and created an issue for this, but in the meantime, we're safe to merge this if we like it enough.

Fixes #169

### DIFF
--- a/src/scripts/clipperUI/clipper.tsx
+++ b/src/scripts/clipperUI/clipper.tsx
@@ -246,7 +246,7 @@ class ClipperClass extends ComponentBase<ClipperState, {}> {
 				this.state.setState({
 					fullPageResult: {
 						data: {
-							failureMessage: Localization.getLocalizedString("WebClipper.Preview.FullPageModeGenericError")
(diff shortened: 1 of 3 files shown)


leakymemory (approved):
This works for me.

VishaalK:
I think @iconix brought up a point that maybe we shouldn't even show the preview if no full page content was found or show some a message similar to the panel we show when local file access is not allowed. This works for me though!

VishaalK (approved):

PR #184, opened 2016-11-01
OptionsPanel tests, fix clip button enablement render in PDF mode, cleanup

### LINKED ISSUE #183: clipButtonEnabled() returns wrong values for PDF mode
Encountered this when I was writing UTs:

1) All Pages selected, something invalid was typed into selection text box
Expected: true, but was false
2) localFilesAllowed is a deceiving name, should be renamed to isLocalFileAndNotAllowed or similar

mttwc:
Added a bunch of UTs for the rendering of an enabled/disabled clip button in optionsPanel for each of our modes. While I was testing this, I noticed that we had incorrect logic for this very thing, and it's now corrected. Also cleaned up some code.

Fixes #183 and #179


src/scripts/clipperUI/clipperStateUtilities.ts
```diff
@@ -0,0 +1,46 @@
...
+				if (!clipperState.pdfPreviewInfo.isLocalFileAndNotAllowed) {
+					return false;
+				} else if (clipperState.pdfResult.status !== Status.Succeeded) {
+					return false;
+				} else if (clipperState.pdfPreviewInfo.allPages) {
+					return true;
+				}
+				// We know at this point that the status is succeeded and the user has a specific page range, so if
+				// it is unparseable or empty, we disallow the clip
+				let pages = StringUtils.parsePageRange(clipperState.pdfPreviewInfo.selectedPageRange);
```
leakymemory: What about the extra parameter which passes the total number of pages?  Without it we won't be able to validate the upper page limit, no?

leakymemory (approved):
Just the one question, but this looks good!

src/scripts/clipperUI/clipperStateUtilities.ts
mttwc: Oh yeah, that's true. I'll add a change and add a UT for this case.

mttwc:
Resolved Chad's comment by adding back the upper bounds check and added a matching UT.

PR #192, opened 2016-11-09
Add logging for size of PDF images

VishaalK:
Nick on the Collab team brought up a good point that we might be sending up a lot of data since we are potentially converting optimized PDFs into dataUri's. This is to see just how big that gap is. 

Preliminary testing suggests the average dataUri is around 3x as big as the average pdf page, which is significant. I have another branch where I am working on throttling the image and finding just the right amount to throttle it to so that the image quality is still acceptable, and the data we are sending up isn't massive. It's a little involved so I split this off and hopefully we can ship it and get 1 weeks worth of data to see where we are at.

### DIFF
--- a/src/scripts/arrayUtils.ts
+++ b/src/scripts/arrayUtils.ts
@@ -10,17 +10,9 @@ export module ArrayUtils {
 		}
 
 		// Calculate the smallest divisor where the largest bucket is size <= maxPerBucket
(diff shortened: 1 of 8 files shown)


leakymemory (approved):

mttwc (approved):
Nice. This will likely conflict with my refactors, but I'll sort that out :)

PR #195, opened 2016-11-10
Split up and test SaveToOneNote (and other cleanup stuff)

### LINKED ISSUE #124: Split up and add UTs for SaveToOneNote.ts
This file is getting pretty big as we made more flexibility to each of our clipping modes. We should look to split this apart to make things easier to validate and understand.

mttwc:
In a nutshell, saving to OneNote was complicated and was not being tested. This change represents the breakup of concerns. An overview of this change would be to look at startClip() in clipper.tsx, which has been greatly shrunk.

There are quite a number of changes to logging where I feel it made sense (details below). This means we'll need to change or add Aria dashboards as appropriate to take advantage of this.

- **Broke up SaveToOneNote.ts into classes that implement clipping, logging, and page creation as seperate concerns**
- Added UTs (mostly for SaveToOneNote) and fixed an issue where we don't reject on the pdf clip if the check for page existance after the initial createPage fails all retries
- **Made use of ONAPI 1.0.6 to implement decorators for our api object to add logging functionality and retry logic. We should no longer have to implement these manually for the api.**
  - **All our api calls in SaveToOneNote now have retry logic. We should see less failures in Aria.**

- Reworked a lot of logging with the refactor. In the past we had log events that would get hit in all clip modes, which meant we couldn't tell them apart, and certain custom properties would be present on some and not on others, depending on the mode, which would make things hard to keep track of.
  - **Broke up clip logging into 2 events**
    - ClipCommonOptions: gets logged regardless of clip mode and has properties relevant to all modes
    - Clip specific events e.g., ClipRegionOptions that have properties relevant to that clip mode (e.g., number of images)

- **With this change, we introduce some example code of using sinon fakeServers to make available more than one endpoint. We can use this to test methods that hit multiple remote endpoints.**

Other:
- Simplified showRatingsPrompt to use primitive values instead of SVs in ClipperState
- Cleaned up unused event custom properties and labels
  - In some cases, consolodated duplicates

Fixes #124


mttwc:
Following up with my talk with @VishaalK I'll try and clean up the promise logic :)

PR #199, opened 2016-11-14
Split up utils.ts

### LINKED ISSUE #99: Utils.ts should be refactored
Utils.ts is an awkward dumping ground for methods that don't belong anywhere else. It also references a bunch of random things like ClipperState etc. That means that code that use Utils.ts to do something really simple will inadvertently import a bunch of other js files. We should split this out in a way that makes sense.

mttwc:
Fixes #99


src/scripts/urlUtils.ts
```diff
@@ -0,0 +1,143 @@
...
+
+	/**
+	 * Gets the query value of the given url and key.
+	 *
+	 * @param url The URL to get the query value from
+	 * @param key The query key in the URL to get the query value from
+	 * @return Undefined if the key does not exist; "" if the key exists but has no matching
+	 * value; otherwise the query value
+	 */
+	export function getQueryValue(url: string, key: string): string {
```
leakymemory, line 58: We should probably replace these methods with either [qs](https://www.npmjs.com/package/qs) or [query-string](https://www.npmjs.com/package/query-string).

leakymemory (approved):
Thanks for cleaning this up.

src/scripts/urlUtils.ts
mttwc: Will track in another issue

PR #201, opened 2016-11-14
Use _.assign and _.extend in our code base

### LINKED ISSUE #147: Since we have lodash as a dependency now, we can replace a lot of our utility methods with it
Things like `arrayUtils.ts` and `createUpdatedObject` in `utils.ts`. Would also help in #99

mttwc:
Fixes #147 

#NLOC


leakymemory (approved):
This is exactly what I was looking for over the weekend!

PR #207, opened 2016-11-16
Finish removing the serverroot and example stuff

larsendaniel:
Most of this was removed a while ago, the only thing left is the tests, which
I've moved more appropriately to /target/tests


leakymemory (approved):
Thanks for this Daniel!

mttwc:
@larsendaniel do you know which files are hitting the 5s timeout? We've recently started faking out setTimeout and sinon fakeTimers so I'm wondering if we missed something.

larsendaniel:
No idea  :|

mttwc:
If you open up your index.html (tests), it will display all the runtimes on the right. That might help with diagnosing.

larsendaniel:
It's the TOTAL time that was hitting 5 seconds, not a single test.  I'm not sure why that was causing it to abort.  I'm not sure if the change to merge all the tests together is what caused it or not

PR #208, opened 2016-11-16
Bundle tests into a single file

larsendaniel:
This improves the BundleTests task from about 40.0 seconds down to about
1.4 seconds.  :D


mttwc:
This is awesome! Our main concern is that now that all the js files are concatenated, and our test files use globals, we're no longer sandboxing modules. This morning I was making a change on my end to convert all the test modules to use classes (and therefore use state instead of global vars), so when I'm done, I'll integrate with your changes and they will become true sandboxes again :)

mttwc:
Also noticed that there's an additional benefit - if the require (or import) has a bad path, we will get a compiler warning. This is far better than mistyping the wrong test file name into tests.html and not getting any warnings of any sort :D

larsendaniel:
Yeah, I had the same thought when I did this.  Putting each test into some kind of module would resolve that issue.

larsendaniel:
Heh, yeah, it helped me notice that things weren't quite lined up.  There were a few extra, and a few missing.   We could almost generate this file from the actual directory contents now

mttwc:
Good call! I'll put up a PR for the sandbox-ifying of tests as soon as its done and I'll rope you in for that :P

mttwc:
@leakymemory I think we can merge this in now. My changes will come later today.

PR #212, opened 2016-11-16
Organize tests into classes

### LINKED ISSUE #213: Fix and reintegrate commented out tests

mttwc:
Refactor tests to minimize use of globals by using classes. They should also look much neater because of clearer formatting.

Also Fixes #213


mttwc:
@larsendaniel feel free to chip in your 2c!

mttwc:
After lots of investigating with browserify caching, it's actually creating more overhead with all the files being written, I will remove those changes.

larsendaniel:
Yup, this looks right to me.  👍

larsendaniel:
Hmmm, actually I noticed that we still need the following blocks that I had to put in.  `protected afterEach() {
        Settings.setSettingsJsonForTesting(undefined);
    }`

So the tests can still potentially affect each other.  But this is still a good fix.  :shipit:

mttwc:
Yep, that's because we use static state in a lot of places, which is a little unfortunate. I can perhaps move some of this into the base abstract class so we don't have to remember to do it.

leakymemory (approved):

Issue #221, opened 2016-11-17
Refactor tabbing tests to get rid of duplicate code


mttwc:
We're doing the same thing each time ... looping through the objects and checking tabIndexes. We can move all this repeat code into a more central place. 

#GottaGetThatNLOC

PR #222, opened 2016-11-17
Clean up HelperFunctions, test initialization, and test sandboxing

### LINKED ISSUE #193: Clean up HelperFunctions.ts
This was originally used to set up test environments for component testing, but since then, we have shoved every helper we can think of in there. This should be broken up.

mttwc:
1) Broke up HelperFunctions into smaller files that make sense
2) Move global test initialization logic to tests.ts
3) Reset the front end storage on every test. The idea is that TestModule.ts will worry about resetting global state so that the developer doesn't have to. This will allow the tests to be far more sandboxed in nature, removing the need to manually clear storage in certain test files if modified in any way.
4) Initialized 'probably necessary' global state before each test, this way you don't have to worry about hitting undefined errors due to dependencies on global objects that have not been defined.

In the future, **I suggest that we avoid using static state/classes as much as possible**. In an object-oriented world, we really should be using objects so we can make our test-lives easier through mocking, programming to the interface, etc. In the future, we may even want to consider using a DI framework.

Fixes #193


mttwc:
Was going to work on https://github.com/OneNoteDev/WebClipper/issues/221 in a different branch but accidentally pushed it to remote in this one. I'll work on that here.

leakymemory (approved):
Cleaning the house feels good! :-)

PR #226, opened 2016-11-21
Fixed broken pdf viewport test, minor cleanup of ObjectUtils

mttwc:
In a nutshell, for very tiny pages, since the spinner is larger than the page viewport, it stretches the viewport to larger than what it should be. Fixed by adding max-height and max-width on the page viewport, and enforcing the spinner itself to be no larger than the viewport.


VishaalK:
Great!

PR #227, opened 2016-11-22
Port pdf header tests to pdf options tests, styling cleanup, prop/state logic simplification, added more tests

mttwc:


VishaalK (approved):
dope

PR #228, opened 2016-11-22
Relax range invalidation

mttwc:
We now allow commas that have whitespace between them. We simply ignore them and keep parsing.


VishaalK:
Looks good!

PR #235, opened 2016-11-23
Fix no preview on invocation in non-article/non-pdf

mttwc:
The if-else logic in this branch could return undefined. Fixed it so that it always returns a value.


mttwc:
Just going to merge this in because it's an important fix and no one's in the office :P

PR #236, opened 2016-11-23
Fix overflowing content during animations

### LINKED ISSUE #233: During animations, it's possible for content to overflow the container during the transition
I have this fixed on one of my branches. You simply need to set overflow: none; to the culprits. This should make our animations look a little cleaner!

mttwc:
Fixes #233


mttwc:
Merging this as it's important and no one's in the office

PR #240, opened 2016-11-23
Simple saveablefactory tests

mttwc:
Converted oneNoteSaveableFactory to be an object rather than using static
Added some simple test scenarios to make sure we don't throw exceptions or reject when given legit input


VishaalK:
Zero to testable in 5 seconds flat. Legendary!

🐑

PR #242, opened 2016-11-23
Underline 'More' on PDF options. Added 'Less' for when the options are expanded.

mttwc:
Just my suggested styling changes. We can check with the designer to see if he likes it.


VishaalK:
Less is more is less...is more. 

I like it!

mttwc:
Localizing the string on my end.

PR #243, opened 2016-11-28
Use Popper.js for Popover

VishaalK:
Initially, I wanted to render a popover as a `div` within the `pdfClipOptions.tsx` component. However, nesting the element here means it cannot overflow outside of the `optionsPanel` content, so the workaround is to insert the popover at the end of body and position it correctly.

`popper.js` is a library that does that but I initially avoided because I thought it was overkill. However, it is actually really powerful and could fuel something like an FRE or a call-out experience for the Clipper, so I think it is okay to take it on as a dependency. 

A better way to do this would be to wrap the `popper` in a component and be able to include that, but its not as easy as it sounds (although there is a gist out there giving some pointers). 

@mttwc This works for me and passes the tests, but I am concerned I am using `config` wrong. I pass the popover handler to the `input` tag. Can you double check this logic?

Edit: Confirmed the styling works on Safari


src/scripts/clipperUI/clipper.tsx
```diff
@@ -620,7 +622,18 @@ class ClipperClass extends ComponentBase<ClipperState, {}> {
 		return signOutState;
 	}
 
-	private handleStartClip() {
+	private handleStartClip(): void {
+		const pdfPreviewInfo = this.state.pdfPreviewInfo;
+		if (this.state.currentMode.get() === ClipMode.Pdf && !pdfPreviewInfo.allPages) {
```
mttwc: Consider adding another sanity check to make sure pdfResult is success.

src/scripts/stringUtils.ts
```diff
@@ -1,22 +1,35 @@
 import {ObjectUtils} from "./objectUtils";
 import {Localization} from "./localization/localization";
 
+import {Status} from "./clipperUI/status";
+
 import * as _ from "lodash";
 
 export module StringUtils {
+	export interface ParsedPageRange {
+		status: Status;
```
mttwc: Consider using a different enum instead of Status (which we reserve for async operations). The reason for this is because from a dev standpoint, for a synchronous function, it's hard to tell what Success or Failed etc. is. There will also be Statuses like InProgress that aren't going to be applicable.

src/scripts/stringUtils.ts
```diff
@@ -1,22 +1,35 @@
...
+		result: any;
```
mttwc: You should be able to use a union type here to make typing stronger.

mttwc:
Additional comments:
- Popover is incorrectly overlaid above the section picker dropdown when it is opened (z-index issue)
- Popover still shows when All Pages is selected (we can resolve this by consolodating state to a single source of truth - see comments below)

src/scripts/clipperUI/clipperStateUtilities.ts
```diff
@@ -19,11 +20,13 @@ export module ClipperStateUtilities {
 					return false;
 				} else if (clipperState.pdfPreviewInfo.allPages) {
 					return true;
+				} else if (!clipperState.pdfPreviewInfo.allPages && ObjectUtils.isNullOrUndefined(clipperState.pdfPreviewInfo.selectedPageRange)) {
```
mttwc, line 23: First check is not necessary, but hmmm, it does make it readable. I'll leave it to you if you want to leave it in there or not :)

src/scripts/clipperUI/components/pdfClipOptions.tsx
```diff
@@ -96,30 +106,62 @@ class PdfClipOptionsClass extends ComponentBase<PdfClipOptionsState, ClipperStat
...
 
-	private invalidPageRange(): boolean {
-		let pagesToShow = StringUtils.parsePageRange(this.props.clipperState.pdfPreviewInfo.selectedPageRange);
-		let validUpperBounds = this.props.clipperState.pdfResult.status === Status.Succeeded ? _.every(pagesToShow, (ind: number) => {
-			return ind <= this.props.clipperState.pdfResult.data.get().pdf.numPages();
-		}) : true;
+	// Destroy the currently existing popover
+	// Then create a new one if necessary
+	// TODO: make this into an actual component
+	private handlePopoverLifeCycle() {
```
mttwc: Right now we're using a static field to maintain the popover, and we expose a lot of low-level methods such as destroy() which can be scary. Additionally, I don't think PdfClipOptions should have to worry about popover lifecycles (admittedly, we can definitely improve our MVC-logic throughout our codebase :P ). I recommend a couple of things:

a) Wrap the popover object in a class that implements an interface that exposes some high-level methods. Example methods could be show(), hide(), setText() etc.
b) Have a popover manager class where you can pass in a string key and get back the popover object associated with it. Creation of popovers can also be done through this class. That way, we can ensure that we don't have duplicate popovers, and at the same time, we can scale popovers to multiple use-cases in the future, all while keeping things testable.

src/scripts/clipperUI/components/pdfClipOptions.tsx
```diff
@@ -96,30 +106,62 @@ class PdfClipOptionsClass extends ComponentBase<PdfClipOptionsState, ClipperStat
...
+				removeOnDestroy: true
+			});
+		}
+	}
+
+	private getErrorMessageForInvalidPageRange(): string {
+		const pdfPreviewInfo = this.props.clipperState.pdfPreviewInfo;
+		let parsePageRangeOperation = StringUtils.parsePageRange(pdfPreviewInfo.selectedPageRange, this.props.clipperState.pdfResult.data.get().pdf.numPages());
+		if (parsePageRangeOperation.status === Status.Succeeded) {
+			throw Error("Given that shouldShowPopover is true, parsing the pageRange should never succeeded: PageRange: " + pdfPreviewInfo.selectedPageRange);
```
mttwc: nit: should say 'succeed'

src/scripts/clipperUI/components/previewViewer/pdfPreview.tsx
```diff
@@ -161,9 +159,12 @@ class PdfPreviewClass extends PreviewComponentBase<PdfPreviewState, ClipperState
 		let pdfResult = this.props.clipperState.pdfResult.data.get();
 
 		// Determine which pages should be marked as selected vs unselected
-		let pagesToShow = StringUtils.parsePageRange(this.props.clipperState.pdfPreviewInfo.selectedPageRange);
-		if (!pagesToShow) {
+		let pagesToShow;
```
mttwc: Similar to one of the above comments, we should make use of strong typing.

src/scripts/constants.ts
```diff
@@ -161,10 +165,12 @@ export module Constants {
 		export var pdfHeaderTitle = "pdfHeaderTitle";
 		export var pageRangeControl = "pageRangeControl";
 
+		// pdfClipOptions
 		export var checkboxToDistributePages = "checkboxToDistributePages";
 		export var checkboxToAttachPdfDisabled = "checkboxToAttachPdfDisabled";
 		export var checkboxToAttachPdf = "checkboxToAttachPdf";
 		export var moreClipOptions = "moreClipOptions";
+		export var popover = "popover";
```
mttwc: #popover doesn't look like it's being used to style anything. Do we still need it as an id? I'd rather this be just a class so we can have better re-use.

src/scripts/constants.ts
```diff
@@ -20,6 +20,10 @@ export module Constants {
 		export var textAreaInput = "textAreaInput";
 		export var textAreaInputMirror = "textAreaInputMirror";
 
+		// optionsPanel
```
mttwc: Consider commenting this with "popover" or similar, instead of "optionsPanel", as we are likely going to re-use this in future experiences.

src/scripts/saveToOneNote/oneNoteSaveableFactory.ts
```diff
@@ -172,6 +173,12 @@ export class OneNoteSaveableFactory {
...
 	private static getPageIndicesToSendInPdfMode(clipperState: ClipperState) {
 		let pdf = clipperState.pdfResult.data.get().pdf;
-		return clipperState.pdfPreviewInfo.allPages ? _.range(pdf.numPages()) : StringUtils.parsePageRange(clipperState.pdfPreviewInfo.selectedPageRange, pdf.numPages()).map(value => value - 1);
+		const parsePageRangeOperation = StringUtils.parsePageRange(clipperState.pdfPreviewInfo.selectedPageRange, pdf.numPages());
+		if (parsePageRangeOperation.status !== Status.Succeeded) {
+			throw new Error("User is clipping an invalid page range");
+		}
+		const pageRange = parsePageRangeOperation.result as number[];
+
+		return clipperState.pdfPreviewInfo.allPages ? _.range(pdf.numPages()) : pageRange.map(value => value - 1);
```
mttwc: Check allPages and return the _.range(pdf.numPages()) at the beginning so you don't need to compute the pageRange result every time.

src/scripts/stringUtils.ts
```diff
@@ -34,35 +47,57 @@ export module StringUtils {
...
 				}
 				valueToAppend = _.range(lhs, rhs);
 			} else {
 				// The currentValue is not a single digit or a valid range
-				return undefined;
+				return {
+					status: Status.Failed,
+					result: currentValue
+				};
+				// return undefined;
```
mttwc: nit: remove comment

src/strings.json
```diff
@@ -103,7 +103,7 @@
 	"WebClipper.Preview.Header.AttachPdfFile": "Attach PDF File",
 	"WebClipper.Preview.Header.PdfTooLargeToAttach": "PDF too large to attach",
 	"WebClipper.Preview.Header.PdfDistributePagesCheckboxLabel": "Create page for each PDF page",
-	"WebClipper.Preview.Header.PdfInvalidPageRange": "Invalid page range.",
+	"WebClipper.Preview.Header.PdfInvalidPageRange": "We couldn\u0027t find page",
```
mttwc: I think @leakymemory suggested a while back doing something like "We couldn\u0027t find page {0}" where {0} is replaced with the page number. I'm not sure if all localized strings will keep the page number at the end of the string, so this is our safest bet.

mttwc (changes requested):
A lot of my comments relate to code-cleanliness, and not functionality. I'm ok with merging this in as long to as we are actively tracking places where we can improve our code :)

src/scripts/stringUtils.ts
VishaalK: I originally had a different enum called `OperationResult` which was just `Succeeded`, `Failed`. That is a fair point about `InProgress`. It should be quick enough converting it back.

src/scripts/stringUtils.ts
VishaalK: Good idea!

src/scripts/stringUtils.ts
VishaalK: Done

src/scripts/clipperUI/components/previewViewer/pdfPreview.tsx
VishaalK: Done

src/scripts/clipperUI/components/pdfClipOptions.tsx
VishaalK: Done

src/scripts/constants.ts
VishaalK: Good catch!

src/scripts/constants.ts
VishaalK: Good catch! Was a holdover from when popover was rendering as a child element.

src/scripts/saveToOneNote/oneNoteSaveableFactory.ts
VishaalK: Done

src/scripts/clipperUI/clipper.tsx
VishaalK: Done

src/scripts/clipperUI/clipper.tsx
```diff
@@ -98,7 +99,8 @@ class ClipperClass extends ComponentBase<ClipperState, {}> {
 				isLocalFileAndNotAllowed: true,
 				selectedPageRange: "",
 				shouldAttachPdf: false,
-				shouldDistributePages: false
+				shouldDistributePages: false,
+				shouldShowPopover: false
```
mttwc, line 103: `shouldShowPopover` should also take into account allPages, which is already state. Perhaps we can narrow the scope of `shouldShowPopover` to keep it seperate? I'm brainstorming this now ... it's tough to think of what we should scope this to :P

src/strings.json
VishaalK: Done!

src/scripts/clipperUI/clipper.tsx
```diff
@@ -620,7 +622,18 @@ class ClipperClass extends ComponentBase<ClipperState, {}> {
 		return signOutState;
 	}
 
-	private handleStartClip() {
+	private handleStartClip(): void {
+		const pdfPreviewInfo = this.state.pdfPreviewInfo;
+		if (this.state.currentMode.get() === ClipMode.Pdf && !pdfPreviewInfo.allPages) {
+			const parsePageRangeOperation = StringUtils.parsePageRange(pdfPreviewInfo.selectedPageRange, this.state.pdfResult.data.get().pdf.numPages());
```
mttwc, line 628: Should we move this logic lower to OptionsPanelClass? In OptionsPanelClass we have the onStartClip prop. We can wrap it with this gating logic before passing down the new function to the clipButton enableInvoke.

src/scripts/clipperUI/clipperStateUtilities.ts
VishaalK: Leaving it for now

src/scripts/clipperUI/clipper.tsx
VishaalK: This is a tough problem that I think we should punt and track with an issue. I think `shouldShowPopover` isn't an accurate name. The state variable actually reflects if the user clicked on clip while there was an invalid page range.

src/scripts/clipperUI/clipper.tsx
VishaalK: That seems fine with me. It makes conceptual sense as well. The logic should live in the lowest component it can. The top level `Clipper` knows how to save the Clip, but it is the interaction between the clipButton and the options panel that gates it, not necessarily something at the `Clipper` level.

mttwc:
Regarding the use of config to render the popover:

a) I don't think we need to put this on the config. The config is simply a function that gets called AFTER the element gets rendered, and we usually do this if we want to access the element itself, as it wouldn't be available before the first render. We can likely call this function in the render and it will behave the same.

b) This will probably come a result of your refactor to the popover object, but if possible, we should avoid destroying and replacing the popover every render/config. Once you do your refactor of the popover object, I'll be happy to have a look to see how we can optimize this.

VishaalK:
1) The popover depends on  `document.getElementById(...)`, so wouldn't we want to wait until that is rendered before rendering a popover? It also makes the component way difficult -.-. I will move the function to render() and see if that changes anything.
2) In the Popper.js documentation I didn't see an easy/obvious way of changing an already existing popover, so I think we should stick the destroy/create paradigm until further investigation.

mttwc:
Ok sure. We could perhaps have some caching logic where we only destroy when we need to 'change' the popover. Right now, destroying and recreating on every render is causing a flicker whenever I trigger another render (e.g., expanding/hiding the More options)

mttwc:
Bug: For page ranges involving large numbers, the number in the popover is expressed with an exponential: 

e.g.: 
`parseInt("111111111111111111111111111", 10);`
`1.111111111111111e+26`

mttwc:
@VishaalK be sure to include an entry in THIRD-PARTY-NOTICES.txt :)

src/strings.json
```diff
@@ -103,7 +103,7 @@
 	"WebClipper.Preview.Header.AttachPdfFile": "Attach PDF File",
 	"WebClipper.Preview.Header.PdfTooLargeToAttach": "PDF too large to attach",
 	"WebClipper.Preview.Header.PdfDistributePagesCheckboxLabel": "Create page for each PDF page",
-	"WebClipper.Preview.Header.PdfInvalidPageRange": "Invalid page range.",
+	"WebClipper.Preview.Header.PdfInvalidPageRange": "We couldn\u0027t find page '{0}'",
```
mttwc, line 107: Should use a different key

mttwc:
Consolidating changes into https://github.com/OneNoteDev/WebClipper/pull/249 and closing this.

PR #245, opened 2016-11-28
Fix exponentiation on popup, as well as other parsing failures

mttwc:
Fixed scientific notation showing up on the popover if there's large numbers involved.
Return a failure as soon as possible before attempting to parse HUGE ranges to avoid crashes/out of memory issues.
Fix issue where user is able to clip something like ",,,,,,,,," or ""


VishaalK:
Very thorough. Thanks for fixing the test descriptions too!

PR #246, opened 2016-11-28
[Bug] Convert Popover to a Component

VishaalK:
Through mithril's `config` attribute, we can hook into the lifecycle of a mithril component. The properties `onElementDraw` and on `onElementFirstDraw` in componentBase do this. 

On creation, I create the popover. On update, I update the popover. On unload, I destroy the popover. 

Since it is wrapped in a component, these lifecycle methods are called appropriately when the component itself is removed from the DOM. 

It's not truly the "React way" because the `Popper.js` library is manipulating the DOM, but now the lifecycles match.


src/scripts/clipperUI/components/popper.tsx
```diff
@@ -0,0 +1,73 @@
...
+	content?: string;
+	classNames?: string[];
+	arrowClassNames?: string[];
+	modifiersIgnored?: string[]; // TODO: use a union type of allowed values
+	removeOnDestroy?: boolean;
+
+	id?: string;
+}
+
+// ref element
```
mttwc: nit: remove comment

mttwc (approved):
This looks great! Played around with this on my end and it looks like the teardown is being called in the right places.

src/scripts/clipperUI/components/popper.tsx
```diff
@@ -0,0 +1,73 @@
...
+		return (
+			<div
+				config={this.handlePopoverLifecycle.bind(this)}
+				id={this.props.id} />
+		);
+	}
+}
+
+let component = PopperClass.componentize();
+export { component as Popover };
```
mttwc: We should either name the export component as Popper, or rename the file and above class name to PopoverClass

mttwc:
Looks like the animation logic was not ported over. I'll work on that.

leakymemory (approved):
Looks good.  Don't forget the third-party-notices.txt

mttwc:
Closing this and merging this into a different branch than bug/change-invalid-page-popover

FezVrasta:
@VishaalK and @mttwc 

you can easily disable the `applyStyle` modifier of Popper.js and let React manipulate the DOM.

https://github.com/FezVrasta/popper.js/tree/master#reactjs-and-emberjs-integration

PR #247, opened 2016-11-29
Extension lifecycle logging

mttwc:
Adds a new required context property to our logging called 'ExtensionLifecycle.Id'. At the beginning of the extension's construction, we assign a random guid to this property, and every worker that gets instantiated from this instance of the extension will share the same ExtensionLifecycle.Id. Asssuming we are comparing 2 sessions, this id will be different if:

a) The 2 sessions came from 2 different browser processes
b) The 2 sessions came from the same browser process, but the extension was refreshed or updated between sessions

Why is this new id useful? Sometimes, our customers complain about an issue, only to get back to us saying that it was resolved after a restart of the browser or similar action. With this id, we can now know for sure what issues persist across extension lifecycles, or persist for only one lifecycle (i.e., an issue involving extension or browser 'state'). If the customer has a particular issue that seems to be transient, we can simply look at the ExtensionLifecycle.Id of the error log, and look at all sessions across that same extension lifecycle to look for more meaningful patterns.


leakymemory (approved):
Very nice.

VishaalK:
Noice

PR #250, opened 2016-11-29
Fix WhatsNew failing due to bad promise logic

mttwc:
When we cleaned up promises, we made a mistake here. Whenever we decided it's time to show the WhatsNew, our resolve would get swallowed, and WhatsNew would not appear. Additionally, we keep attempting to show WhatsNew since the previous attempts failed, leading to large amounts of session logging.


VishaalK:
Good detective work 🕵️

PR #261, opened 2016-12-05
Bubble up O365 sign-in errors

### LINKED ISSUE #85: Authentication errors need to be bubbled up to the user
When the authentication service hits an error, it is generally returned as part of the redirect URL.  We are currently logging this error, but really it should be something that we can bubble up to the user.  This is mostly because most of the time we don't have any control over it and the error is specific to that user.

This is tracking the work to somehow bubble up the error message we get back to the user.

Current (initial) thoughts are to add a "more info" link which opens up a popup with the message (as well as the clipper ID/session ID to help debug.

leakymemory:
Fixes #85 

Authentication errors are still one of the biggest problems we hit with the clipper.  In the case of O365, they are particularly hard to address since often the solution involves provisioning by the admin.  In the past, when an O365 user would hit this kind of a problem, they simply got the message, "We couldn't sign you in. Please try again.", which really wasn't very helpful.

This is a quick feature which will bubble up these O365 errors to the user.  Granted, it's still somewhat confusing as some people may not know what the error really means - but it is certainly better than nothing.  Right now, when users hit these kind of errors and they are not able to sign in, they either give up or post a message on the forums.  If they reach out to us, we need to get their id, look up the log information and then respond with the actual error.  This change gets that information to the user much faster.

Since this is hard to test, here is a picture of what it looks like when an O365 sign-in error is detected:
image

And here is what it looks like fully expanded:
image


mttwc:
Will look at the code in a bit. In the meantime, I think from a design standpoint, we should have that 'V' graphic next to "More Information" to indicate that it is interactable.

src/scripts/clipperUI/clipper.tsx
```diff
@@ -589,6 +589,11 @@ class ClipperClass extends ComponentBase<ClipperState, {}> {
 			if (error) {
 				handleSignInEvent.setStatus(Log.Status.Failed);
 				handleSignInEvent.setFailureInfo({ error: error });
+
+				data.errorDescription = error;
```
mttwc: I think you want to use errorObject here in lines 593 and 594 for readability.

src/tests/clipperUI/panels/signInPanel_tests.tsx
```diff
@@ -92,6 +92,50 @@ export class SignInPanelTests extends TestModule {
...
+			strictEqual(document.getElementById(Constants.Ids.signInToggleErrorInformationText).innerText, this.stringsJson["WebClipper.Label.SignInUnsuccessfulMoreInformation"],
+				"The displayed message should be the 'More' message");
+		});
+
+		test("The 'less' button is enabled when a sign-in failure is detected and the more button was clicked (OrgID)", () => {
+			let state = MockProps.getMockClipperState();
+			state.userResult = { status: Status.Failed, data: { lastUpdated: 10000000, updateReason: UpdateReason.SignInAttempt, errorDescription: "OrgId: An error has occured." } };
+			let controllerInstance = MithrilUtils.mountToFixture(<SignInPanel clipperState={state} onSignInInvoked={this.mockSignInPanelProps.onSignInInvoked} />);
+
+			// Simulate that the 'More' button was clicked.
```
mttwc: You could add a test using MithrilUtils.simulateAction() and a .click() call on the relevant element.

src/tests/clipperUI/panels/signInPanel_tests.tsx
```diff
@@ -92,6 +92,50 @@ export class SignInPanelTests extends TestModule {
...
+			ok(!!document.getElementById(Constants.Ids.signInErrorDescription), "The error description is showing");
+		});
+
+		test("The 'more' button is not there when a sign-in failure is detected (MSA)", () => {
+			let state = MockProps.getMockClipperState();
+			state.userResult = { status: Status.Failed, data: { lastUpdated: 10000000, updateReason: UpdateReason.SignInAttempt, errorDescription: "MSA: An error has occured." } };
+
+			let controllerInstance = MithrilUtils.mountToFixture(<SignInPanel clipperState={state} onSignInInvoked={this.mockSignInPanelProps.onSignInInvoked} />);
+
+			ok(!document.getElementById(Constants.Ids.signInToggleErrorInformationText), "The error information should be missing.");
```
mttwc: Comment should say something like 'error information toggler' or similar for clarity.

src/tests/clipperUI/panels/signInPanel_tests.tsx
```diff
@@ -92,6 +92,50 @@ export class SignInPanelTests extends TestModule {
...
+		test("The error description is showing when a sign-in failure is detected and the more button was clicked (OrgID)", () => {
+			let state = MockProps.getMockClipperState();
+			state.userResult = { status: Status.Failed, data: { lastUpdated: 10000000, updateReason: UpdateReason.SignInAttempt, errorDescription: "OrgId: An error has occured." } };
+			let controllerInstance = MithrilUtils.mountToFixture(<SignInPanel clipperState={state} onSignInInvoked={this.mockSignInPanelProps.onSignInInvoked} />);
+
+			// Simulate that the 'More' button was clicked.
+			controllerInstance.setState({ errorDescriptionShowing: true });
+			m.redraw(true);
+
+			ok(!!document.getElementById(Constants.Ids.signInErrorDescription), "The error description is showing");
```
mttwc, line 132: Should add a test to ensure the signInErrorDescription element is not rendered by default (i.e., when the informaiton is collapsed)

src/styles/mainClassDefinitions.less
```diff
@@ -174,6 +174,40 @@
 	padding-bottom: 28px;
 }
 
+#signInErrorDescription {
```
mttwc: You'll want to put these id css definitions in another file since this one's for class definitions. On a side note we should look at the way we're organizing our less files and see if there's a neater way.

src/scripts/clipperUI/panels/signInPanel.tsx
```diff
@@ -60,15 +84,43 @@ class SignInPanelClass extends ComponentBase<{}, SignInPanelProps> {
 							{signInDescription}
 						</span>
 					</div>
-					<a id={Constants.Ids.signInButtonMsa} {...this.enableInvoke(this.onSignInMsa, 1)}>
+					{signInFailureDetected
```
mttwc: Consider moving each of these chunks into a method for readability.
simple example

```
<div>
   {this.getElemOne()}
   {this.getElemTwo()}
   {this.getElemThree()}
</div>
```

mttwc (changes requested):
Nice tests!

mttwc (approved):

mttwc:
Looks good!

mttwc:
I wonder about smaller screen sizes. We're currently expanding to show the message in the main controller, which can potentially can get cut off (I'm thinking about @iconix's laptop for example). Still, this is a very good first step and I think we should ship this asap.

src/tests/clipperUI/panels/signInPanel_tests.tsx
```diff
@@ -92,6 +92,54 @@ export class SignInPanelTests extends TestModule {
 
 			strictEqual(type, AuthType.OrgId, "The OrgId auth type should be recorded");
 		});
+
+		test("The 'more' button is enabled when a sign-in failure is detected (OrgID)", () => {
```
mttwc, line 97: Looks like we don't have any tab tests for this panel. Can you please add them? :)

mttwc (changes requested):

src/tests/clipperUI/panels/signInPanel_tests.tsx
leakymemory: Done.

PR #262, opened 2016-12-05
Update version to 3.3.0

leakymemory:
With the upcoming updates to the PDF mode, let's bump the version to 3.3.0


mttwc:
As discussed let's merge this into @VishaalK's pdf v2 branch. We first need to find out which branch he prefers this to be merged into :P

Issue #265, opened 2016-12-06
Augmentation API is not consistently adding the data-original-src to iframe element for video


mttwc:
Repro: http://www.seattletimes.com/business/amazon/amazoncom-unveils-self-driving-brick-and-mortar-convenience-store/ (article mode)

ONApi expects iframe videos to have a data-original-src attribute on the iframe for each of our videos, and it must not have any query params. In the case of this link (which has a youtube video on it), Augmentation API is incorrectly sending back an iframe *without* the attribute. When the user clips this, the video is not rendered in the page.

mttwc:
Fixed in #266

PR #266, opened 2016-12-06
Add the ability to save videos in the selection clip

### LINKED ISSUE #263: Selection mode does not grab 'allowed' iframe elements such as Youtube videos

mttwc:
In selection mode, we would get rid of all iframes. Now, we check each iframe to see if it's a video we support, then we construct the iframe accordingly so it meets the standards expected by ONApi (i.e., an appropriate data-original-src with no query parameters)

Fixes #263


mttwc:
With the latest commit, also fixes #265 by converting every returned iframe into a video iframe that is compliant for the ONApi clip action. This way, we shouldn't get into a situation where a video returned by AugmentationApi is actually unclippable.

VishaalK:
Looks great!

PR #267, opened 2016-12-06
[Feature] Progress UI and Synchronous Page Create for PDF Mode

VishaalK:
When a user wants to spread their PDF over multiple OneNote pages, they can now select this option in the mainController.

We then sequentially createPages one by one until we go through all the pages that the user wants to save. If any of the create page calls fail, we keep going on. The sequential logic is almost identical to our batching and patching logic.

To give the user some feedback, after each createPage call, the clipping panel will indicate the progress of the clip by saying "Clipping page {0} of {1}...". The way we do this is by passing a callback into `SaveToOneNote.save(saveOptions)`, and `SaveToOneNote` will call this callback with the number of items it has completed, and the total number of items it needs to be completed, so the UI can render it as it sees fit. 

The page range now has a popover that comes up if the page range is invalid.

I've confirmed everything works in RTL (prior to localizing some strings).

### DIFF
--- a/src/scripts/clipperUI/clipper.tsx
+++ b/src/scripts/clipperUI/clipper.tsx
@@ -667,7 +667,7 @@ class ClipperClass extends ComponentBase<ClipperState, {}> {
 			let saveOptions: SaveToOneNoteOptions = {
 				page: saveable,
 				saveLocation: this.state.saveLocation,
-				progressCallback: this.updateProgress.bind(this)
+				progressCallback: this.updateClipSaveProgress.bind(this)
 			};
 
 			let saveToOneNote = new SaveToOneNote(this.state.userResult.data.user.accessToken);
@@ -691,7 +691,7 @@ class ClipperClass extends ComponentBase<ClipperState, {}> {
 		});
 	}
 
-	private updateProgress(numItemsCompleted: number, numItemsTotal: number): void {
+	private updateClipSaveProgress(numItemsCompleted: number, numItemsTotal: number): void {
 		this.state.setState({
 			clipSaveStatus: {
 				numItemsCompleted: numItemsCompleted,
--- a/src/scripts/clipperUI/clipper.tsx
+++ b/src/scripts/clipperUI/clipper.tsx
@@ -662,8 +662,10 @@ class ClipperClass extends ComponentBase<ClipperState, {}> {
 		(new OneNoteSaveableFactory(this.state)).getSaveable().then((saveable) => {
 			let saveOptions: SaveToOneNoteOptions = {
 				page: saveable,
-				saveLocation: this.state.saveLocation
+				saveLocation: this.state.saveLocation,
+				progressCallback: this.updateProgress.bind(this)
 			};
+
 			let saveToOneNote = new SaveToOneNote(this.state.userResult.data.user.accessToken);
 			saveToOneNote.save(saveOptions).then((responsePackage: OneNoteApi.ResponsePackage<any>) => {
 				let createPageResponse = Array.isArray(responsePackage) ? responsePackage[0] : responsePackage;
@@ -683,6 +685,13 @@ class ClipperClass extends ComponentBase<ClipperState, {}> {
 				Clipper.logger.logEvent(clipEvent);
 			});
 		});
+	}
+
+	private updateProgress(numPagesCompleted: number, totalPages: number): void {
+		_.assign(_.extend(this.state.pdfSaveStatus, {
+			numPagesCompleted: numPagesCompleted,
+			totalPages: totalPages
+		}), this.state.setState);
 	}
 
 	private storeLastClippedInformation() {
--- a/src/scripts/clipperUI/clipper.tsx
+++ b/src/scripts/clipperUI/clipper.tsx
@@ -102,7 +102,7 @@ class ClipperClass extends ComponentBase<ClipperState, {}> {
 				shouldDistributePages: false,
 				shouldShowPopover: false
 			},
-			pdfSaveStatus: {
+			clipSaveStatus: {
 				numItemsTotal: undefined,
 				numItemsCompleted: undefined
 			},
@@ -693,7 +693,7 @@ class ClipperClass extends ComponentBase<ClipperState, {}> {
 
 	private updateProgress(numItemsCompleted: number, numItemsTotal: number): void {
 		this.state.setState({
-			pdfSaveStatus: {
+			clipSaveStatus: {
 				numItemsCompleted: numItemsCompleted,
 				numItemsTotal: numItemsTotal
 			}
--- a/src/scripts/clipperUI/clipper.tsx
+++ b/src/scripts/clipperUI/clipper.tsx
@@ -692,14 +692,12 @@ class ClipperClass extends ComponentBase<ClipperState, {}> {
 	}
 
 	private updateProgress(numPagesCompleted: number, totalPages: number): void {
-		console.log("updateProgress start");
 		this.state.setState({
 			pdfSaveStatus: {
 				numPagesCompleted: numPagesCompleted,
 				totalPages: totalPages
 			}
 		});
-		console.log("updateProgress end");
 	}
 
 	private storeLastClippedInformation() {
@@ -750,7 +748,6 @@ class ClipperClass extends ComponentBase<ClipperState, {}> {
 			undefined;
 		let regionSelectorItem = ClipperClass.shouldShowRegionSelector(this.state) ? <RegionSelector clipperState={this.state} /> : undefined;
 		let mainControllerStyle = ClipperClass.shouldShowMainController(this.state) ? { } : { display: "none" };
-		console.log("rendar");
 
 		return (
 			<div>
--- a/src/scripts/clipperUI/clipper.tsx
+++ b/src/scripts/clipperUI/clipper.tsx
@@ -103,8 +103,8 @@ class ClipperClass extends ComponentBase<ClipperState, {}> {
 				shouldShowPopover: false
 			},
 			pdfSaveStatus: {
-				numPagesCompleted: undefined,
-				totalPages: undefined
+				numItemsTotal: undefined,
+				numItemsCompleted: undefined
 			},
 
 			reset: () => {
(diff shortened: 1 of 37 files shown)


src/scripts/clipperUI/clipper.tsx
```diff
@@ -662,8 +666,10 @@ class ClipperClass extends ComponentBase<ClipperState, {}> {
 		(new OneNoteSaveableFactory(this.state)).getSaveable().then((saveable) => {
 			let saveOptions: SaveToOneNoteOptions = {
 				page: saveable,
-				saveLocation: this.state.saveLocation
+				saveLocation: this.state.saveLocation,
+				progressCallback: this.updateProgress.bind(this)
```
VishaalK: Since we don't pass `clipperState` into `SaveToOneNote::save`, I opted to keep the progressUpdate as a callback, but it is easy enough to change.

src/scripts/clipperUI/components/pdfClipOptions.tsx
```diff
@@ -191,8 +198,14 @@ class PdfClipOptionsClass extends ComponentBase<PdfClipOptionsState, ClipperStat
 		);
 	}
 
-	getEnabledAttachmentCheckbox(): any {
+	getAttachPdfCheckbox(clickable: boolean): any {
+		return clickable ? this.getEnabledAttachmentCheckbox() : this.getDisabledAttachmentCheckbox();
+	}
+
+	// TODO: Programmaticaly remove the enableInvoke so that we can collapse the attachmentCheckbox functions into one
```
VishaalK: I split the functions for returning the enabled and disabled versions of the `Attach PDF` checkbox because I didn't know how to programmatically remove the `enableInvoke` that allows the user to interact with the checkbox.

src/scripts/clipperUI/clipper.tsx
mttwc: I think what you did here is fair :)

src/scripts/clipperUI/clipper.tsx
```diff
@@ -685,6 +691,15 @@ class ClipperClass extends ComponentBase<ClipperState, {}> {
 		});
 	}
 
+	private updateProgress(numItemsCompleted: number, numItemsTotal: number): void {
```
mttwc: Method could be more descriptive e.g. 'updatePdfClipProgress'

src/scripts/clipperUI/components/pdfClipOptions.tsx
```diff
@@ -176,13 +177,19 @@ class PdfClipOptionsClass extends ComponentBase<PdfClipOptionsState, ClipperStat
 		const pdfIsTooLarge = pdfHasSucceeded && this.props.clipperState.pdfResult.data.get().byteLength > Constants.Settings.maximumMimeSizeLimit;
 		const disableCheckbox = pdfIsTooLarge || !pdfHasSucceeded;
 
-		return disableCheckbox ? this.getDisabledAttachmentCheckbox() : this.getEnabledAttachmentCheckbox();
+		if (pdfIsTooLarge) {
+			return this.getAttachmentIsTooLargeCheckbox();
+		} else {
```
mttwc: No need for else

src/scripts/clipperUI/components/pdfClipOptions.tsx
mttwc: You could do something like this to cut down on the repeat code. It's not perfect since it doesn't look like you can assign an already-spread object to a variable
```
getAttachPdfCheckbox(enabled: boolean): any {
	const disabledClassName = enabled ? "" : " disabled";
	let pdfPreviewInfo = this.props.clipperState.pdfPreviewInfo;

	let inner = (
		<div>
			<div className={"pdf-indicator pdf-checkbox-indicator" + disabledClassName}></div>
			{enabled && pdfPreviewInfo.shouldAttachPdf ? <div className={Constants.Classes.checkboxCheck}></div> : undefined}
			<div className="pdf-label-margin">
				<span className={"pdf-label" + disabledClassName}>{Localization.getLocalizedString("WebClipper.Label.AttachPdfFile") + " "}
					<span className={"sub-label" + disabledClassName}>{Localization.getLocalizedString("WebClipper.Label.AttachPdfFileSubText")}</span>
				</span>
			</div>
		</div>
	);

	if (enabled) {
		return <div className="pdf-control" id={Constants.Ids.checkboxToAttachPdf} {...this.enableInvoke(this.onCheckboxChange, 66, !pdfPreviewInfo.shouldAttachPdf) }>
			{inner}
		</div>;
	}
	return <div className={"pdf-control" + disabledClassName} id={Constants.Ids.checkboxToAttachPdf} tabIndex={66}>
		{inner}
	</div>;
}
```

mttwc: Another thing to consider would be to simply pass in undefined into the enableInvoke call. If you go that route, I'd modify enableInvoke to call logClickEvent only if handleMethod is not undefined. This way, you can still set a tabIndex through enableInvoke without specifying any interactivity. This way the above solution can be even simpler.

src/scripts/clipperUI/components/popover.tsx
```diff
@@ -1,5 +1,6 @@
 import {Constants} from "../../constants";
 import {StringUtils} from "../../stringUtils";
+import {Localization} from "../../localization/localization";
```
mttwc: Remove import

src/scripts/clipperUI/mainController.tsx
```diff
@@ -260,8 +261,14 @@ export class MainControllerClass extends ComponentBase<MainControllerState, Main
 				return <RegionSelectingPanel clipperState={this.props.clipperState} />;
 			case PanelType.ClippingToApi:
 				if (this.props.clipperState.currentMode.get() === ClipMode.Pdf) {
-					return <ClippingPanelWithDelayedMessage clipperState={this.props.clipperState}
-						delay={Constants.Settings.pdfClippingMessageDelay} message={Localization.getLocalizedString("WebClipper.ClipType.Pdf.ProgressLabelDelay")} />;
+					if (this.props.clipperState.pdfPreviewInfo.shouldDistributePages) {
+						return <ClippingPanelWithProgressIndicator clipperState= { this.props.clipperState } />;
+					} else {
```
mttwc: No need for else

src/scripts/clipperUI/pdfSaveStatus.ts
```diff
@@ -0,0 +1,4 @@
+export interface PdfSaveStatus {
```
mttwc: Name could be generalized for reuse in case we want to do something similar with other modes in the future.

src/scripts/saveToOneNote/oneNoteSaveablePdfSynchronousBatched.ts
```diff
@@ -0,0 +1,65 @@
...
+	/**
+	 * @index starts at 0 and refers to the pages already in the saveable, NOT THE PAGE OF THE PDF ITSELF
+	 */
+	public getPage(index?: number): Promise<OneNoteApi.OneNotePage> {
+		if (ObjectUtils.isNullOrUndefined(index) || index === 0) {
+			// They are asking for the first page
+			return Promise.resolve(this.page);
+		}
+
+		const pageNumber = this.pageIndexes[index - 1];
```
mttwc, line 39: I'd move the above comment here for clarity

src/scripts/saveToOneNote/saveToOneNote.ts
```diff
@@ -37,16 +44,61 @@ export class SaveToOneNote {
...
-		if (options.page.getNumPatches() > 0) {
-			return this.rejectIfNoPatchPermissions(options.saveLocation).then(() => {
-				return this.saveWithoutCheckingPatchPermissions(options).then((responsePackage) => {
-					// TODO: unnecessary then
+		if (options.page.getNumPages() > 1) {
+			return this.saveMultiplePagesSynchronously(options);
+		} else {
+			if (options.page.getNumPatches() > 0) {
+				return this.rejectIfNoPatchPermissions(options.saveLocation).then(() => {
+					return this.saveWithoutCheckingPatchPermissions(options).then((responsePackage) => {
```
mttwc: This TODO could be resolved

src/scripts/saveToOneNote/oneNoteSaveablePdfSynchronousBatched.ts
leakymemory, line 39: I think a line or two of comment here will help explain _why_ we need to subtract 1.  At first glance, it's not clear.

mttwc (changes requested):

leakymemory (approved):
Very nice!

src/scripts/clipperUI/pdfSaveStatus.ts
VishaalK: Good point!

src/scripts/clipperUI/clipper.tsx
VishaalK: Thanks!

src/scripts/clipperUI/components/pdfClipOptions.tsx
VishaalK: Good ideas! I will implement them and then verify that they work. I am a bit scared that virtualDOM won't pick up attaching/detaching of handlers if I do this, but this will be a good chance to find out.

src/scripts/saveToOneNote/saveToOneNote.ts
VishaalK: Done!

src/scripts/saveToOneNote/oneNoteSaveablePdfSynchronousBatched.ts
VishaalK: I added a lengthy comment that I hope is instructive when I inevitably forget why I did this

src/scripts/clipperUI/pdfSaveStatus.ts
VishaalK: Done!

src/scripts/clipperUI/mainController.tsx
VishaalK: Done!

src/scripts/clipperUI/components/popover.tsx
VishaalK: Done!

src/scripts/clipperUI/components/pdfClipOptions.tsx
VishaalK: Done!

PR #269, opened 2016-12-08
[Feature] Change the BATCH pipeline to use the new BatchRequest builder object

VishaalK:
This is the corresponding change for https://github.com/OneNoteDev/OneNoteApi/pull/7.

This code path is right now not used, since we createPages synchronously. However, it is still worth having. Tests will fail when we switch over in 1-2 months, as expected,


leakymemory:
Looks good.

PR #270, opened 2016-12-08
Feature/pdf mode polish

leakymemory:
Minor updates to polish the new options a bit.


src/styles/mainClassDefinitions.less
```diff
@@ -294,6 +293,10 @@
 		user-select: auto;
 		float: left;
 
+		&:focus {
+			color: #fff;
```
leakymemory: It occurs to me that this isn't quite right.  We need to distinguish not only when the input has focus, but also need to know when the radio button is enabled/disabled.  Working on it now.

VishaalK:
Update:

1. The mode that is NOT selected is now slightly faded, to indicate it is not active. 
2. Fixed some spacing issues
3. Moved the invalid page range popover to appear below the page range input, instead of to the right.

PR #273, opened 2016-12-08
[Feature] Add ability for users to split their PDF into multiple OneNote pages

VishaalK:
This PR:

1. Moves PDF options from the header bar to the mainController
2. Adds a new option "New note for each PDF page" that will split the PDF into multiple pages in the users' OneNote
3. Adds a popover to check for invalid page ranges


VishaalK:
Just tested it on all the browsers and it is working.

PR #278, opened 2016-12-21
[Bug] Remove dependency on numeric value of enum for ContentModels in augmentationHelper.ts

### LINKED ISSUE #276: Recipes labeled as Products, Products labeled as Articles
Repro by invoking the clipper on each of these links:

http://www.seriouseats.com/recipes/2016/10/pressure-cooker-tomato-sauce.html

https://www.amazon.com/Logitech-Harmony-Ultimate-Customizable-Cabinet/dp/B00BQ5RY1G/ref=gbph_img_m-3_f3ee_77320560?smid=ATVPDKIKX0DER&pf_rd_p=f8221e3c-958b-4601-81d4-a545a755f3ee&pf_rd_s=merchandised-search-3&pf_rd_t=101&pf_rd_i=11516390011&pf_rd_m=ATVPDKIKX0DER&pf_rd_r=2YRZ878V00MVRHG9GPHM

VishaalK:
Fixes #276 

We added a new ContentModel, `EntityKnowledge`, on the server and this shifted our ENUM values. 

Recipes were now 4 instead instead of 3, so Recipes were showing up as Products. Products were now 5 instead of 5, so they were converted to `Screenshot`, which is unsupported, so they were defaulting to Article.

There is a work-item for the server to return ContentModel as a string, so that when they add new ContentModels, this doesn't break us.

### DIFF
--- a/src/scripts/contentCapture/augmentationHelper.ts
+++ b/src/scripts/contentCapture/augmentationHelper.ts
@@ -19,13 +19,14 @@ import * as Log from "../logging/log";
 import {CaptureFailureInfo} from "./captureFailureInfo";
 
 export enum AugmentationModel {
-	None = 0,
-	Article = 1,
-	BizCard = 2,
-	Recipe = 3,
-	Product = 4,
-	Screenshot = 5,
-	Wrapstar = 6
+	None,
+	Article,
+	BizCard,
+	EntityKnowledge,
+	Recipe,
+	Product,
+	Screenshot,
+	Wrapstar
 }
 
 export interface AugmentationResult extends CaptureFailureInfo {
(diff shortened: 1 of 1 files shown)


mttwc:
Nice catch! I can now go back to clipping my recipes in peace!

mttwc:
I'm afraid of what might happen if the server changes the enum such that the number that it sends down is out of range. We should add some try-catch logic for this case.

EDIT: Spoke to @VishaalK about this last week. We already have safe fallback logic (i.e., goes to Article mode).

mttwc:
I just realized that there will be quite a few people who don't autoupdate and they will be broken by:
a) Changes in the enum number being sent down
b) The change to a string being sent down
I think we need to preserve as much old server-side behavior as possible ...

leakymemory (approved):

PR #280, opened 2016-12-21
[Bug] Turn on Firefox Tooltips

VishaalK:
There was a bug with Firefox Tooltips where they weren't rendered correctly. Instead of fixing it, I let it simmer, and it fixed itself :). There will be a big deluge of tooltips in the beginning, and it might give us a nice idea of how big our install base on Firefox is. 

I am leaving this PR open so we can just merge it in whenever want to turn these on.


mttwc:
Was this tested on Nightly/Prod Firefox/both?

VishaalK:
I double-checked that it worked on Nightly.

It doesn't work when I merge it with the latest version of master, but it DOES work when I leave it as is. Very frustrating :(

VishaalK:
Matthew and I couldn't repro the issue on Nightly so I am saying we are good to go

mttwc:
^ confirmed. This looks fine.

jogonzal:
Closing as this is old and not merged

PR #282, opened 2016-12-28
Feature/unwriteable cookie detect

### LINKED ISSUE #12: Show a better error message when cookies are blocked by the browser
Because we still rely on the browser's ability to write cookies on the www.onenote.com domain, there is a possibility that another extension, or the browser itself if certain options are set, can block that cookie write.  We should try to detect this situation and let the user know how to get around it.  Nearly always the answer is to add an exception for https://www.onenote.com.

The challenge here is that the message itself will need to be somewhat generic as we will have no idea what is actually blocking the cookie.  Hopefully we can write in a way that the user can just "figure it out".

leakymemory:
Fixes #12 

Ok, one of the big sign-in issues that people are hitting right now is related to cookies being blocked in the browser.  This is generally because either they have a setting enabled in their browser which blocks third-party cookies, or they have an extension installed which effectively blocks our cookie from getting written.

This change is combined with a server-side change where we are effectively _testing_ to see if cookies can be written on the onenote.com domain, and if not, show a message to the user letting them know why they are unable to sign in.

Essentially, this is how the flow works.  As part of the sign-in process, there is a redirect back to the /auth endpoint.  As part of that response we request a cookie be written.  The next call we make, the POST request which retrieves the user's information tries to pass up that cookie.  As part of that response, there is a new property ```cookieInRequest```, which lets us know whether or not the cookie was passed up along with the request.  If the sign in failed and this property is set to false then we can show an appropriate message which will (hopefully) help the user fix the issue.


src/scripts/clipperUI/panels/signInPanel.tsx
```diff
@@ -60,52 +60,84 @@ class SignInPanelClass extends ComponentBase<SignInPanelState, SignInPanelProps>
 			&& this.props.clipperState.userResult.data.errorDescription.indexOf("OrgId") === 0;
 	}
 
-	errorDescriptionControlHandler() {
-		this.setState({ errorDescriptionShowing: !this.state.errorDescriptionShowing });
+	signInFailureThirdPartyCookiesBlocked(): boolean {
```
mttwc, line 67: There's some shared 'checks' between this method and signInFailureContainsErrorDescription that can be moved to another function. i.e., `!!this.props.clipperState.userResult && !!this.props.clipperState.userResult.data && this.props.clipperState.userResult.data.updateReason === UpdateReason.SignInAttempt`

src/scripts/clipperUI/panels/signInPanel.tsx
```diff
@@ -60,52 +60,84 @@ class SignInPanelClass extends ComponentBase<SignInPanelState, SignInPanelProps>
...
-				<a id={Constants.Ids.signInErrorMoreInformation} {...this.enableInvoke(this.errorDescriptionControlHandler, 10) }>
+	errorMoreInformationTogggle() {
+		if (this.signInErrorDetected()) {
+			return <div id="signInErrorToggleInformation">
+				<a id={Constants.Ids.signInErrorMoreInformation} {...this.enableInvoke(this.debugInformationControlHandler, 10) }>
 					<img id={Constants.Ids.signInToggleErrorDropdownArrow} src={ExtensionUtils.getImageResourceUrl("dropdown_arrow.png")} />
 					<span id={Constants.Ids.signInToggleErrorInformationText}
 						style={Localization.getFontFamilyAsStyle(Localization.FontFamily.Light)}>
-						{this.state.errorDescriptionShowing
+						{this.state.debugInformationShowing
```
mttwc, line 87: A More/Less information panel could potentially be moved and tested within its own component.

src/scripts/constants.ts
```diff
@@ -345,7 +348,6 @@ export module Constants {
 	}
 
 	export module Urls {
-		// export var serviceDomain = "https://minint-per7o4f.redmond.corp.microsoft.com";
```
mttwc, line 348: Thanks for cleaning up :)

src/scripts/extensions/authenticationHelper.ts
```diff
@@ -67,18 +67,23 @@ export class AuthenticationHelper {
...
+					this.user.set({ updateReason: updateReason, writeableCookies: writeableCookies });
 				}
 
 				resolve(this.user.get());
 			}, (error: OneNoteApi.GenericError) => {
 				getInfoEvent.setStatus(Log.Status.Failed);
 				getInfoEvent.setFailureInfo(error);
+
 				this.user.set({ updateReason: updateReason });
 				resolve(this.user.get());
```
mttwc: On a side note: This line repeats itself in the good and bad case, so we can consolidate it into the final then() block.

src/scripts/extensions/authenticationHelper.ts
```diff
@@ -67,18 +67,23 @@ export class AuthenticationHelper {
 			let getInfoEvent: Log.Event.PromiseEvent = new Log.Event.PromiseEvent(Log.Event.Label.GetExistingUserInformation);
 			getInfoEvent.setCustomProperty(Log.PropertyName.Custom.UserInformationStored, !!storedUserInformation);
 			this.clipperData.getFreshValue(ClipperStorageKeys.userInformation, getUserInformationFunction, updateInterval).then((response: TimeStampedData) => {
-				getInfoEvent.setCustomProperty(Log.PropertyName.Custom.FreshUserInfoAvailable, !!response);
```
mttwc, line 70: Side note on line 55: should be collapsable to this
```
let getUserInformationFunction = () => {
	return this.getClipperInfoCookie(clipperId).then((cookie) => {
		return this.retrieveUserInformation(clipperId, cookie);
	});
};

File: src/tests/extensions/authenticationHelper_tests.ts [hunk 6]
```diff
@@ -187,71 +188,7 @@ export class AuthenticationHelperSinonTests extends TestModule {
...
-				fullName: undefined
-			};
-			this.server.respondWith(
-				"POST", this.authUrl,
-				[200, { "Content-Type": "application/json" },
-				JSON.stringify(invalidUserInformation)
-			]);
-
-			this.authentationHelper.retrieveUserInformation(this.clipperId).then((responsePackage) => {
-				strictEqual(responsePackage.parsedResponse, undefined, "The response should be undefined");
```
mttwc, line 254: Thanks for updating UTs!

mttwc (approved):
Nice! Looking forward to seeing the server-side stuff!

src/scripts/clipperUI/panels/signInPanel.tsx
leakymemory: I'm going to hold off on this one for now.  I can see this being something we use in other places at some point, and so I'll wait until then to do the refactoring.

src/scripts/extensions/authenticationHelper.ts
leakymemory: _So_ much cleaner.  Thanks!

src/scripts/extensions/authenticationHelper.ts
leakymemory: Good call.

PR #288, opened 2016-12-30
Fix the regular expression for onenote.com

### LINKED ISSUE #275: What's New can incorrectly appear in the sign in popup

leakymemory:
Fixes #275 

It was possible for the What's New dialog to appear in the sign-in dialog.  This is because the
regular expression used to detect onenote.com was a bit over-aggressive and the redirect URL on
the sign-in's query parameter list was triggering it.

The fix is to scope the regular expression a little such that it only matches on the real page.


mttwc:
Nice catch!

PR #289, opened 2017-01-03
If the user token expires mid-patch clip, refresh it

### LINKED ISSUE #225: When clipping a HUGE pdf, the user token can expire during this time
This results in us saving an incomplete (PATCHed) pdf to OneNoteAPI.

mttwc:
Fixes #225

### DIFF
--- a/src/scripts/saveToOneNote/saveToOneNote.ts
+++ b/src/scripts/saveToOneNote/saveToOneNote.ts
@@ -2,6 +2,7 @@ import {Constants} from "../constants";
 import {PromiseUtils} from "../promiseUtils";
 import {Settings} from "../settings";
 import {StringUtils} from "../stringUtils";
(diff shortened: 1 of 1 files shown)


src/scripts/saveToOneNote/saveToOneNote.ts
```diff
@@ -179,19 +180,45 @@ export class SaveToOneNote {
 
 						Promise.all([getRevisionsPromise, timeoutPromise]).then((values) => {
 							let revisions = values[0] as OneNoteApi.Revision[];
-							this.getApi().updatePage(pageId, revisions).then(() => {
+
+							let thenCb = () => {
```
mttwc, line 184: To avoid repeat code I pulled out the callbacks, but this still feels messy. I'm open to improvements if any of you have suggestions :)

leakymemory (approved):

PR #292, opened 2017-01-04
Fix the embedded YouTube regular expression

### LINKED ISSUE #291: Not cleaning out all of the query parameters for embedded videos
It looks like we aren't fully cleaning out the query parameters on some embedded videos:

http://www.macrumors.com/2017/01/03/faraday-future-ff-91-reveal/

This is the ONML that is getting sent:
<iframe width="600" height="338" src="https://www.youtube.com/embed/1Y8hUf0Np6k?start=900" data-original-src="https://www.youtube.com/watch?v=1Y8hUf0Np6k?start=900"></iframe>

Removing the "?start=900" and re-issuing in Fiddler fixes the problem.

leakymemory:
Fixes #291 

Update the embedded YouTube regular expression such that it does not match any query parameters that might be there.


mttwc:
Looks good

PR #297, opened 2017-01-11
Bug/o365 sign in errors

### LINKED ISSUE #296: No longer showing O365 sign-in errors
It looks like there was a regression in the way we show the O365 sign-in errors with the last change.

leakymemory:
Fixes #296 

The details in each commit describe the fixes more completely, but in general, the O365 errors were not being shown because the updateReason property was not being set in the returning error object.  Because of this, the logic around when to show the message to the user was broken.

I also made a related UI fix and updated the string to include the live.com domain as a needed exception.

### DIFF
--- a/src/scripts/clipperUI/clipper.tsx
+++ b/src/scripts/clipperUI/clipper.tsx
@@ -576,11 +576,18 @@ class ClipperClass extends ComponentBase<ClipperState, {}> {
 		let handleSignInEvent = new Log.Event.PromiseEvent(Log.Event.Label.HandleSignInEvent);
 
 		this.setState({ userResult: { status: Status.InProgress } });
-		type ErrorObject = { updateReason: UpdateReason, correlationId?: string, error: string, errorDescription: string };
+
+		type ErrorObject = {
+			updateReason: UpdateReason,
+			correlationId?: string,
+			error: string,
+			errorDescription: string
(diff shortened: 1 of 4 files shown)


VishaalK:
After cleaning up the `if (!updatedUser) {...}` logic it looks good to go!

src/scripts/clipperUI/clipper.tsx
```diff
@@ -576,41 +576,36 @@ class ClipperClass extends ComponentBase<ClipperState, {}> {
 		let handleSignInEvent = new Log.Event.PromiseEvent(Log.Event.Label.HandleSignInEvent);
 
 		this.setState({ userResult: { status: Status.InProgress } });
-		type ErrorObject = { correlationId?: string, error: string, errorDescription: string };
-		Clipper.getExtensionCommunicator().callRemoteFunction(Constants.FunctionKeys.signInUser, { param: authType, callback: (data: UserInfo | ErrorObject) => {
+		type ErrorObject = { updateReason: UpdateReason, correlationId?: string, error: string, errorDescription: string };
```
VishaalK: nit: can we declare the type so it looks more like an object instead of all on one line

VishaalK:
LGTM

mttwc:
Looks good!

PR #303, opened 2017-01-18
Only inject the install marker on onenote online

### LINKED ISSUE #166: A user reported that the div we insert on pageNav can cause some problems with other add-ons
We insert a div on pageNav in order to detect if the Clipper is already installed to make some experiences smoother.

image

A user reported it might cause issues so we are going to investigate so we don't break other extensions

mttwc:
Fixes #166 

Admittedly this is something I should have done from the start >_<


src/scripts/extensions/appendIsInstalledMarker.ts
```diff
@@ -1,9 +1,13 @@
 let oneNoteWebClipperInstallMarkerClassName = "oneNoteWebClipperIsInstalledOnThisBrowser";
 let marker = document.createElement("DIV");
```
VishaalK: Use lowercase 'div' instead of upper case 'div'

VishaalK:
Looks good!

PR #305, opened 2017-01-19
[Bug] Fix Region mode being off on high dpi screens like Retina or Surface Book in Firefox

### LINKED ISSUE #257: Firefox, OSX12: Region selection is off
It looks like in Firefox, region selection is off and what the user selects is definitely not what they're getting.

Verified with most recent version of Firefox (50.0.2)

VishaalK:
We confirmed that the region screenshot mode on Retina's (and now Surface Pro 4) in Firefox was returning images that were wildly off.

I did some debugging and digging, and while the image returned by `captureVisibleTab` seems to be the same dataUrl, using this dataUrl as a source for `canvas.drawImage(...)` would have different results because of browser implementation. It has something to do with how the browser keeps the images in memory on high dpi screens in what is called a "backing store". 

I found a workaround for detecting this and getting the screenshot, but the resolution of the screenshot is lower, and this bug is currently being tracked here: https://bugzilla.mozilla.org/show_bug.cgi?id=1278507. The good news is the bug looks active :) 

There is a more heavyweight work-around here: http://outof.me/chrome-extension-retina-capturevisibletab-translate3d-2-x-res/

All of this info will go in the new issue I will open about the resolution of the screenshot

Fixes #257


src/scripts/clipperUI/regionSelector.tsx
```diff
@@ -199,8 +200,17 @@ class RegionSelectorClass extends ComponentBase<RegionSelectorState, ClipperStat
...
+		const query = "(-webkit-min-device-pixel-ratio: 2), (min-device-pixel-ratio: 2), (min-resolution: 192dpi)";
+		const isHighDpiScreen = matchMedia(query).matches;
+		const isFirefoxWithHighDpiDisplay = this.props.clipperState.clientInfo.clipperType === ClientType.FirefoxExtension && isHighDpiScreen;
+
 		// Firefox reports this value incorrectly if this iframe is hidden, so store it now since we know we're visible
-		this.devicePixelRatio = window.devicePixelRatio;
+		// In addition to this, Firefox currently has a bug where they are not using devicePixelRatio correctly 
+		// on HighDPI screens such as Retina screens or the Surface Pro 4
+		// Bug link: https://bugzilla.mozilla.org/show_bug.cgi?id=1278507 
+		this.devicePixelRatio = isFirefoxWithHighDpiDisplay ? window.devicePixelRatio / 2 : window.devicePixelRatio;
```
mttwc, line 213: When FF fixes this, would this value be incorrect?

VishaalK: Yes, it will be.

I hardcoded the first and second points used in the region capture and Firefox and Chrome returned different results, which seems the bug is in the implementation, and there doesn't seem to be a way to collapse the logic to use the "correct" values once FF fixes the bug.

VishaalK: On the plus side, since I am on the bug thread, when they fix it I will get an e-mail and it will be some time before it makes to Firefox Prod, since it has to go through Nightly an Dev editions first.

leakymemory (approved):

PR #315, opened 2017-02-14
Safe updates of the packages we depend on

leakymemory:
Quick iteration over our packages and updating those that are safe to update right now.


larsendaniel:
👍

leakymemory:
@larsendaniel It's interesting, there is a new Mithril package (1.0.1), but all the unit tests break with it (and I suspect the product as well).  I'm going to dig in at some point and see what's going on.

PR #316, opened 2017-02-16
Switch the order for article detection.

leakymemory:
We had things a little out of order, where it was possible to have a PDF,
but still put things in article mode by default (if the URL happened to
match the heuristic). Switched the order so that the other modes will get
priority as appropriate.


mttwc:
Looks good!

PR #325, opened 2017-02-24
[Bug] Add ARIA to augmentation header controls

VishaalK:
Adding role="button" makes the screen reader interpret these elements as interactable, and then they read the "aria-label". Aria-pressed indicates if they are toggled or not. The aria-label does not affect users who don't use the screen readers.

In process of localizing...so hold off on reviewing for a bit.


VishaalK:
Not sure if I should localize this. Gonna ping Harrie

PR #328, opened 2017-02-25
[Bug] Add screen reader feedback when user selects a mode

VishaalK:
When a user changes mode, there is no screen reader indicator to tell the user the mode successfully changed.

The recommended way to get around this is to use `aria-live`. When this attribute is put on an HTML Element, the screen reader will read whatever is inside whenever it changes. In this case, we can put the current mode into that html element, and render it along with the mode button selector.

Right now, it renders twice, but that is still better than not saying anything.


leakymemory:
Why is this rendering twice?

leakymemory (approved):

PR #343, opened 2017-03-17
Add UpdateReason to getexistinguserinfo log

### LINKED ISSUE #342: WriteableCookies on GetExistingUserInformation event is misleading
This value is only accurate on the sign in attempts, as the service only checks for writeable cookies then. Because of this, the false vs true ratio is going to be super bloated. We'll need to add a change to only log this put on sign ins. Luckily, one of the method's arguments is the UpdateReason, so we can filter on that :)

mttwc:
Fixes #342. Currently, we log WriteableCookies every time we fetch user info. This can be misleading as e.g. it would always return false if the user isn't signed-in on initial Clipper invocation. After this we'll be able to more easily monitor actual success/failure rates (i.e., on sign ins only).


iconix (approved):

VishaalK (approved):

leakymemory:
The dashboard matching the monitor has been created as well.

PR #350, opened 2017-03-29
Remove local/extension URLs from the DOM

### LINKED ISSUE #348: We need to strip out injected extension URLs in the DOM before sending to augmentation service
It looks like there are some extensions which inject CSS or code into the DOM of web pages, which will have a URL reference to the extension itself.  When we pass these references up to the server to render the full page screenshot, or retrieve the article, these references can "throw off" the rendering.

We need to approach this from both directions.  On the client side we should remove these invalid URLs before sending them up, and on the server side, we need to do a better job of ignoring these invalid URLs.  By far the cheapest of these is the client-side work, so we should do this quickly and unblock the majority of the users who will be in this situation.

leakymemory:
Fixes #348 

Because we are sending up the DOM to be rendered on the service (for the
full page screenshot), we sometimes run into problems because of a URL
that is impossible to resolve, URLs like ones injected by another
extension.  This change basically strips out those URLs before the DOM is
sent up.

Oh, and because I really wanted to make a private method, I converted
DomUtils.ts to a true class, which had a little bit of a cascading effect.
I have a bigger change that I want to do which converts all of our
exported functions into classes, but that's for another day.

### DIFF
--- a/src/scripts/domParsers/domUtils.ts
+++ b/src/scripts/domParsers/domUtils.ts
@@ -2,160 +2,171 @@
 // they rename `sanitize` to `sanitizeHtml`, which is the name of the actually exported function
 declare let sanitizeHtml;
 
+// This private has technically been deprecated, but is present in all major browsers as of 2/12/14 and offers
+// the best perf for the UTF byte manipulation we need for truncation and byte size estimation.
+// DO NOT USE IT WITHOUT CHECKING IF IT EXISTS -- in case browser vendors actually remove them
+// ReSharper disable once InconsistentNaming
+declare function unescape(s: string): string;
+
 import {Constants} from "../constants";
 import {ObjectUtils} from "../objectUtils";
-import {VideoUtils} from "./videoUtils";
+import {SupportedVideoDomains, VideoUtils} from "./videoUtils";
 
 import {VideoExtractorFactory} from "./VideoExtractorFactory";
+
+export interface EmbeddedVideoIFrameSrcs {
+	srcAttribute: string;
+	dataOriginalSrcAttribute: string;
+}
 
 /**
  * Dom specific Helper utility methods
  */
-export module DomUtils {
-	export module Tags {
-		export const a = "a";
-		export const b = "b";
-		export const applet = "applet";
-		export const article = "article";
-		export const audio = "audio";
-		export const base = "base";
-		export const body = "body";
-		export const br = "br";
-		export const button = "button";
-		export const canvas = "canvas";
-		export const center = "center";
-		export const cite = "cite";
-		export const del = "del";
-		export const div = "div";
-		export const em = "em";
-		export const embed = "embed";
-		export const figure = "figure";
-		export const font = "font";
-		export const h1 = "h1";
-		export const h2 = "h2";
-		export const h3 = "h3";
-		export const h4 = "h4";
-		export const h5 = "h5";
-		export const h6 = "h6";
-		export const head = "head";
-		export const header = "header";
-		export const hr = "hr";
-		export const html = "html";
-		export const i = "i";
-		export const iframe = "iframe";
-		export const img = "img";
-		export const input = "input";
-		export const li = "li";
-		export const link = "link";
-		export const main = "main";
-		export const map = "map";
-		export const menu = "menu";
-		export const menuitem = "menuitem";
-		export const meta = "meta";
-		export const meter = "meter";
-		export const noscript = "noscript";
-		export const object = "object";
-		export const ol = "ol";
-		export const p = "p";
-		export const progress = "progress";
-		export const script = "script";
-		export const span = "span";
-		export const source = "source";
-		export const strike = "strike";
-		export const strong = "strong";
-		export const style = "style";
-		export const sub = "sub";
-		export const sup = "sup";
-		export const svg = "svg";
-		export const table = "table";
-		export const td = "td";
-		export const title = "title";
-		export const tr = "tr";
-		export const u = "u";
-		export const ul = "ul";
-		export const video = "video";
-	}
+export class DomUtils {
+	public static tags = {
+		a: "a",
+		b: "b",
+		applet: "applet",
+		article: "article",
+		audio: "audio",
+		base: "base",
+		body: "body",
+		br: "br",
+		button: "button",
+		canvas: "canvas",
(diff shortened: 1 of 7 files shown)


src/tests/domParsers/domUtils_tests.ts
```diff
@@ -94,6 +94,27 @@ export class DomUtilsTests extends TestModule {
 				"The non Web Clipper iframe should still be present");
 		});
 
+		test("removeExtensionLinkHrefs should remove all URLs which refer to local extensions, but not any others", () => {
```
mttwc: Sweet!

src/tests/domParsers/domUtils_tests.ts
```diff
@@ -94,6 +94,27 @@ export class DomUtilsTests extends TestModule {
...
+			let linksToCreate: any = [
+				{ id: "chromeExtensionLink", url: "chrome-extension://abc123abc123abc123/test.css", shouldRemain: false},
+				{ id: "edgeExtensionLink", url: "edge-extension://abc123abc123abc123/test.css", shouldRemain: false },
+				{ id: "fileExtensionLink", url: "file://c:/foo/bar/test.css", shouldRemain: false },
+				{ id: "firefoxExtensionLink", url: "firefox-extension://abc123abc123abc123/test.css", shouldRemain: false },
+				{ id: "safarExtensionLink", url: "safari-extension://abc123abc123abc123/test.css", shouldRemain: false },
+				{ id: "normaHttplLink", url: "https://foo/bar/test.css", shouldRemain: true },
```
mttwc: nit: spelling on normalHttpLink

src/tests/domParsers/domUtils_tests.ts
```diff
@@ -94,6 +94,27 @@ export class DomUtilsTests extends TestModule {
 				"The non Web Clipper iframe should still be present");
 		});
 
+		test("removeExtensionLinkHrefs should remove all URLs which refer to local extensions, but not any others", () => {
+			let linksToCreate: any = [
+				{ id: "chromeExtensionLink", url: "chrome-extension://abc123abc123abc123/test.css", shouldRemain: false},
+				{ id: "edgeExtensionLink", url: "edge-extension://abc123abc123abc123/test.css", shouldRemain: false },
+				{ id: "fileExtensionLink", url: "file://c:/foo/bar/test.css", shouldRemain: false },
+				{ id: "firefoxExtensionLink", url: "firefox-extension://abc123abc123abc123/test.css", shouldRemain: false },
+				{ id: "safarExtensionLink", url: "safari-extension://abc123abc123abc123/test.css", shouldRemain: false },
```
mttwc: nit: spelling on safariExtensionLink

src/tests/domParsers/domUtils_tests.ts
```diff
@@ -94,6 +94,27 @@ export class DomUtilsTests extends TestModule {
 				"The non Web Clipper iframe should still be present");
 		});
 
+		test("removeExtensionLinkHrefs should remove all URLs which refer to local extensions, but not any others", () => {
+			let linksToCreate: any = [
+				{ id: "chromeExtensionLink", url: "chrome-extension://abc123abc123abc123/test.css", shouldRemain: false},
```
mttwc: We could add a couple different file extensions in the test

src/tests/domParsers/domUtils_tests.ts
```diff
@@ -94,6 +94,27 @@ export class DomUtilsTests extends TestModule {
...
+			];
+
+			for (let link of linksToCreate) {
+				this.createLinkElementWithHref(link.id, link.url);
+			}
+
+			DomUtils.removeUnsupportedHrefs(document);
+
+			for (let link of linksToCreate) {
+				ok(!!document.getElementById(link.id) === link.shouldRemain, "Url not removed when it was expected to: " + link.url);
```
mttwc: Can be converted to strictEqual for better test result reporting

mttwc (approved):
Nice change, and thanks for doing some code cleanup!

src/tests/domParsers/domUtils_tests.ts
mttwc: Test name should match method name

src/scripts/domParsers/domUtils.ts
```diff
@@ -974,4 +996,13 @@ export module DomUtils {
 
 		return scrollValue * 100;
 	}
+
+	private static isLocalReferenceUrl(url: string): Boolean {
+		return (
+			url.indexOf("chrome-extension://") === 0 ||
```
mttwc: I wonder if we're better off returning false if the url doesn't start with http:// or https://. Also you can loop through the schemes to avoid dupe code.

src/scripts/domParsers/domUtils.ts
```diff
@@ -907,6 +922,13 @@ export module DomUtils {
 			"onfocus",
 			"onblur"
 		];
+
+		// for (let attribute of attributesToRemove) {
```
mttwc: Remove commented-out stuff

src/scripts/domParsers/domUtils.ts
```diff
@@ -2,179 +2,190 @@
...
-		export const table = "table";
-		export const td = "td";
-		export const title = "title";
-		export const tr = "tr";
-		export const u = "u";
-		export const ul = "ul";
-		export const video = "video";
-	}
+export class DomUtils {
+	public static tags = {
```
mttwc, line 26: Side thought: I'm wondering if we should stop using these constants and use hardcoded strings instead. I think constants aren't very useful on immutable values, and it's much more readable if we use hardcoded strings (i.e., the javascript way of doing things).

src/tests/domParsers/domUtils_tests.ts
leakymemory: Yep, late rename and missed this.

src/scripts/domParsers/domUtils.ts
leakymemory: Makes sense to me. I'm honestly not sure why we did this in the first place.

src/scripts/domParsers/domUtils.ts
leakymemory: Hmm, you know, that just makes sense. :-)

PR #357, opened 2017-04-02
Add the <pre> and <code> tags to the SanitizeHtml allowed list

### LINKED ISSUE #352: We're stripping out the <pre> before we clip
It looks like the <pre> tag is being preserved up to the point that we actually clip.  At that point it seems to be stripped out.

leakymemory:
Fixes #352 

When we moved to use SanitizeHtml we pass in a list of tags that are allowed.  We missed a couple.*

*Note that I don't remember if the code tag is supported in the ONML, but I decided to let it through just in a case.


VishaalK:
Looks good!

PR #358, opened 2017-04-04
Remove base64 encoded binaries in <style>

### LINKED ISSUE #349: Some extensions may inject really large files into the page
We recently had one of our users reach out to us because they were unable to clip any page.  The response was always, "No content found".  From the logs it wasn't clear what was going on, but with some help from the user in isolating things, it was discovered that the ViewEdit extension was injecting CSS into every page which included a base64 encoded binary of a custom font-face.  It looked something like this:

`</style><style id="viewedit-css" type="text/css">@font-face{font-family:Open Sans;src:url(data:application/x-font-ttf;base64,AAEAAAATAQAABAAwRFNJR54SRB0AAzucAAAVdEdERUYAJgOvAAM3fAAAAB5HUE9TCzcPNwADN5wAAAA4R1NVQg4rPbcAAzfUAAADxk9TLzKhPp7JAAABuAAAAGBjbWFwKasvaAAAELQAAAQaY3Z0IA9NGKQAAB2QAAAAomZwZ21...`

In this case it well exceeds the size limit for an XHR.  We need to figure out a non-hacky way to strip this kind of thing out of the copied DOM that we send up to the service.

leakymemory:
Fixes #349 

It turns out that there are a lot of exensions in the world, and sometimes those extensions modify the DOM of the page (ours included). Generally we don't have a problem with this, but if an extension injects a large base64-encoded binary into the mix, then we run into problems.  This is mostly because the request size very quickly can be bloated beyond the max.  This change targets one particular place where this happens the most, in the <style> tag where someone may inject a custom font into the page.  When we detect this, we remove the style (from our copy of the DOM) before sending it up.

### DIFF
--- a/src/scripts/domParsers/domUtils.ts
+++ b/src/scripts/domParsers/domUtils.ts
@@ -202,7 +202,7 @@ export class DomUtils {
 	 */
 	public static removeStylesWithBase64EncodedBinaries(doc: Document): void {
 		DomUtils.domReplacer(doc, "style", (node: HTMLElement) => {
-			return node.innerHTML.indexOf("data:application") !== -1 ? document.createElement("style") : node;
+			return node.innerHTML.indexOf("data:application") !== -1 ? undefined : node;
 		});
 	}
 
--- a/src/scripts/domParsers/domUtils.ts
+++ b/src/scripts/domParsers/domUtils.ts
@@ -195,6 +195,17 @@ export class DomUtils {
 		return sanitizedHtml;
 	}
 
+	/**
+	 * Many extensions inject their own stylings into the page, and generally that isn't a problem. But,
+	 * occasionally the styling includes a specific font, which can be very, very large. This method
+	 * removes any base64 encoded binaries defined in any <style> tags.
+	 */
+	public static removeStylesWithBase64EncodedBinaries(doc: Document): void {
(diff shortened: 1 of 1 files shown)


src/scripts/domParsers/domUtils.ts
```diff
@@ -195,6 +195,17 @@ export class DomUtils {
...
 	}
 
+	/**
+	 * Many extensions inject their own stylings into the page, and generally that isn't a problem. But,
+	 * occasionally the styling includes a specific font, which can be very, very large. This method
+	 * removes any base64 encoded binaries defined in any <style> tags.
+	 */
+	public static removeStylesWithBase64EncodedBinaries(doc: Document): void {
+		DomUtils.domReplacer(doc, "style", (node: HTMLElement) => {
+			return node.innerHTML.indexOf("data:application") !== -1 ? document.createElement("style") : node;
```
mttwc: We should be able to return undefined instead of an empty style element to cut down on the HTML we send over the wire.

mttwc (approved):
Looks good! Just one minor consideration.

src/scripts/domParsers/domUtils.ts
leakymemory: Good call.

PR #360, opened 2017-04-05
Replace undefined event variable with `ev`, the properly scoped event

VishaalK:
Apparently lib.d.ts has a line `declare var event: Event | null;` which is maybe why the compiler did not catch this?

I am not sure


mttwc:
:shipit: 
I wonder if we can add a compile time warning to avoid variables of this name.

PR #374, opened 2017-05-19
Feature/Edge Centennial updates

leakymemory:
With the Centennial update of Windows, Edge has two key changes that we have been waiting for:
* They have fixed the problem with the background thread not being able pass along cookies written on the foreground thread.
* They have enabled the CaptureVisibleTab API, which we need to enable the region mode.


mttwc (approved):
Woo! Negative lines of code!

PR #377, opened 2017-07-14
Remove clip anytime in fullpage message

### LINKED ISSUE #376: 'You can clip whenever in Full Page mode' message should no longer display
We have removed this a while back because some sites can't be clipped in full page mode. The message needs to be removed.

mttwc:
Fixes #376


VishaalK (approved):
Looks good! This would be quite annoying for users

PR #379, opened 2017-07-26
Remove dupe correlation ids in retry logic

### LINKED ISSUE #378: Retries should not use the same correlation id
This is making debugging difficult on our end, as it allows for one correlation id to correspond to multiple API requests.

mttwc:
Fixes #378

### DIFF
--- a/src/scripts/saveToOneNote/oneNoteApiWithLogging.ts
+++ b/src/scripts/saveToOneNote/oneNoteApiWithLogging.ts
@@ -1,16 +1,15 @@
 import {Constants} from "../constants";
 
 import {Clipper} from "../clipperUI/frontEndGlobals";
+import {OneNoteApiUtils} from "../clipperUI/OneNoteApiUtils";
 
 import * as Log from "../logging/log";
 
 export class OneNoteApiWithLogging implements OneNoteApi.IOneNoteApi {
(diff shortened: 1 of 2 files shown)


src/scripts/saveToOneNote/oneNoteApiWithLogging.ts
```diff
@@ -90,13 +89,12 @@ export class OneNoteApiWithLogging implements OneNoteApi.IOneNoteApi {
 			let event = new Log.Event.PromiseEvent(eventLabel);
 			let correlationId: string;
 			return func().then((response) => {
+				event.setCustomProperty(Log.PropertyName.Custom.CorrelationId, response.request.getResponseHeader(Constants.HeaderValues.correlationId));
```
jogonzal, line 92: (As I understand this code change, we will no longer set correlationid on the header) If we time out, you can still use sessionid to look for logs, so this is good.

jogonzal (approved):

VishaalK (approved):
Nice job!

src/scripts/saveToOneNote/oneNoteApiWithLogging.ts
mttwc: Yep :)

PR #385, opened 2017-12-13
Fix auth redirect detection by giving benefit of the doubt

mttwc:
Some time a little while ago, our redirect detection became wonky, and it meant that when the sign-in window auto-closes, in rare cases, we detect this as a manual close by the user. As a result of those cases, we intentionally never complete sign-in flow and the user is directed back to the sign-in panel. Now we give the benefit of the doubt and always try and complete sign-in flow, and only determine the reason if and when it fails.

### DIFF
--- a/src/scripts/extensions/extensionWorkerBase.ts
+++ b/src/scripts/extensions/extensionWorkerBase.ts
@@ -499,15 +499,18 @@ export abstract class ExtensionWorkerBase<TTab, TTabIdentifier> {
 
 		this.uiCommunicator.registerFunction(Constants.FunctionKeys.signInUser, (authType: AuthType) => {
 			return this.doSignInAction(authType).then((redirectOccurred) => {
-				if (redirectOccurred) {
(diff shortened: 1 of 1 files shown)


iconix (approved):
Wondering if the code comment could be more detailed... Do you know what changed in the sign-in flow to break this experience, or why redirect detection is wonky/unreliable? Also would suggest providing a date instead of saying "recently" since this code is in maintenance mode.

Does anything change with the normal auth experience with this change?

Otherwise LGTM- although not as familiar with auth parts of this codebase, so not an ideal final approver here :)

mttwc:
Unfortunately, I don't know the inner workings of why redirect detection is wonky, outside of the bool being false when we expected true.

As for a UX perspective, there is one UX change - in extremely rare cases, where the user attempts the E2E auth flow, and something went wrong (user obj is null), and we incorrectly detect that a redirect did not occur, we think this is a 'user cancelled sign-in' state and don't show the error message with correlation-id. However, this is much better than the user signing in successfully but not getting past the sign-in panel :)

VishaalK (approved):

PR #395, opened 2018-05-17
fix tabIndex orders

JaneHazen:


.editorconfig
```diff
@@ -0,0 +1,6 @@
+root = true
```
mttwc, line 1: This should be git rm'd

src/scripts/clipperUI/components/previewViewer/previewComponentBase.tsx
```diff
@@ -75,97 +157,16 @@ export abstract class PreviewComponentBase<TState, TProps extends ClipperStatePr
...
+					<div id={Constants.Ids.previewUrlContainer}>
 						<span aria-label={sourceUrlCitationPrefix}>{sourceUrlCitationPrefix}</span>
-						<a href={sourceUrl} target="_blank" aria-label={sourceUrl} title={sourceUrl}>{sourceUrl}</a>
+						<a tabIndex={-1} href={sourceUrl} target="_blank" aria-label={sourceUrl} title={sourceUrl}>{sourceUrl}</a>
 					</div>
 					: undefined}
 			</div>
 		);
 	}
-
```
mttwc, line 90: Out of curiosity, why were the methods moved around?

mttwc (changes requested):

PR #397, opened 2018-05-17
Add groupings for mode selector buttons

JaneHazen:


src/scripts/clipperUI/components/annotationInput.tsx
```diff
@@ -78,17 +74,21 @@ class AnnotationInputClass extends ComponentBase<AnnotationInputState, ClipperSt
...
+						<span aria-label={Localization.getLocalizedString("WebClipper.Label.AnnotationPlaceholder")}>{Localization.getLocalizedString("WebClipper.Label.AnnotationPlaceholder")} </span>
 					</div>
 				</div>
 			);
 		} else {
 			return (
 				<div id={Constants.Ids.annotationContainer}>
-					<pre id={Constants.Ids.annotationFieldMirror}
-						className={Constants.Classes.textAreaInputMirror}>
+				<pre
```
mttwc: This should be indented

src/scripts/clipperUI/components/modeButton.tsx
```diff
@@ -18,6 +19,21 @@ class ModeButtonClass extends ComponentBase<{}, ModeButtonProps> {
 		this.props.onModeSelected(this.props.myMode);
 	}
 
+	private getAriaPositionInSet(name) {
```
mttwc: We should not use this approach to get the posinset:
- Depending on what kind of page the user clips (e.g., pdf vs a normal web page), different buttons will show up
- From an abstraction standpoint, a mode button is a low level component that should not know about the existence of other mode buttons

src/scripts/clipperUI/components/modeButton.tsx
```diff
@@ -29,12 +45,14 @@ class ModeButtonClass extends ComponentBase<{}, ModeButtonProps> {
...
+				className={className} role="option" aria-setsize="4" aria-posinset={this.getAriaPositionInSet(idName)}
 				id={idName} title={this.props.tooltipText ? this.props.tooltipText : ""}
-				{...this.enableInvoke(this.buttonHandler, this.props.tabIndex) }>
-				<img className="icon" src={this.props.imgSrc} />
+				{...this.enableInvoke(this.buttonHandler, this.props.tabIndex)} tabIndex={this.props.selected ? 40 : ""}
+				aria-selected={this.props.selected}>
+			<img className="icon" src={this.props.imgSrc} />
 				<span className="label buttonLabelFont"
-					style={Localization.getFontFamilyAsStyle(Localization.FontFamily.Regular)}>
+							style={Localization.getFontFamilyAsStyle(Localization.FontFamily.Regular)}>
```
mttwc: Remove indentation

src/scripts/clipperUI/components/modeButton.tsx
```diff
@@ -29,12 +45,14 @@ class ModeButtonClass extends ComponentBase<{}, ModeButtonProps> {
 		let idName: string = clipMode + "Button";
 
 		return (
-			<a className={className} aria-selected={this.props.selected} role="button"
+			<a
+				className={className} role="option" aria-setsize="4" aria-posinset={this.getAriaPositionInSet(idName)}
 				id={idName} title={this.props.tooltipText ? this.props.tooltipText : ""}
-				{...this.enableInvoke(this.buttonHandler, this.props.tabIndex) }>
-				<img className="icon" src={this.props.imgSrc} />
+				{...this.enableInvoke(this.buttonHandler, this.props.tabIndex)} tabIndex={this.props.selected ? 40 : ""}
```
mttwc: this.enableInvoke sets the tabIndex on the element. You should not call enableInvoke and set tabIndex on the same property

mttwc: Add a comment as to why we only set the tabIndex on the button if it's the current selection

src/scripts/clipperUI/components/modeButton.tsx
```diff
@@ -29,12 +45,14 @@ class ModeButtonClass extends ComponentBase<{}, ModeButtonProps> {
 		let idName: string = clipMode + "Button";
 
 		return (
-			<a className={className} aria-selected={this.props.selected} role="button"
+			<a
+				className={className} role="option" aria-setsize="4" aria-posinset={this.getAriaPositionInSet(idName)}
```
mttwc: setsize is not necessarily 4

src/scripts/clipperUI/components/modeButton.tsx
```diff
@@ -1,14 +1,15 @@
...
 import {ClipMode} from "../clipMode";
+import {ComponentBase} from "../componentBase";
 
 export interface ModeButtonProps {
 	imgSrc: string;
 	label: string;
 	myMode: ClipMode;
 	tabIndex: number;
 	selected?: boolean;
+	"aria-posinset": string;
```
mttwc: It looks like these aren't being used below.

src/scripts/clipperUI/components/annotationInput.tsx
```diff
@@ -78,17 +74,21 @@ class AnnotationInputClass extends ComponentBase<AnnotationInputState, ClipperSt
 		if (!this.state.opened) {
 			return (
 				<div id={Constants.Ids.annotationContainer}>
-					<div id={Constants.Ids.annotationPlaceholder} style={Localization.getFontFamilyAsStyle(Localization.FontFamily.Regular)} {...this.enableInvoke(this.handleAnnotateButton, 210)}>
-						<img src={ExtensionUtils.getImageResourceUrl("editorOptions/add_icon_purple.png")} />
-						<span>{Localization.getLocalizedString("WebClipper.Label.AnnotationPlaceholder")}</span>
+					<div
+						id={Constants.Ids.annotationPlaceholder}
+						style={Localization.getFontFamilyAsStyle(Localization.FontFamily.Regular)} {...this.enableInvoke(this.handleAnnotateButton, 210)}
+						role="textbox">
```
mttwc: The role here should not be a textbox as the user can't type into it yet.

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -1,21 +1,65 @@
...
-
 import {ModeButton} from "./modeButton";
 
-class ModeButtonSelectorClass extends ComponentBase<{}, ClipperStateProp> {
+export interface ModeButtonSelectorState {
+	mouseKeyOnButton?: number;
+}
+
+class ModeButtonSelectorClass extends ComponentBase<ModeButtonSelectorState, ClipperStateProp> {
+	constructor(props) {
```
mttwc: This assumes that there's a fixed number of buttons, whereas it could be any number.

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -1,21 +1,65 @@
...
 import {Localization} from "../../localization/localization";
-
 import {ClipMode} from "../clipMode";
 import {ClipperStateProp} from "../clipperState";
 import {ComponentBase} from "../componentBase";
-
 import {ModeButton} from "./modeButton";
 
-class ModeButtonSelectorClass extends ComponentBase<{}, ClipperStateProp> {
+export interface ModeButtonSelectorState {
```
mttwc: ModeButtons should behave in isolation and should have no knowledge of other buttons, abstraction-wise, this should be moved to a component that knows about all ModeButtons

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -1,21 +1,65 @@
...
+		if (selectedID === 0) {
+			selectedID = 4;
+		}
+		this.setState({ mouseKeyOnButton: selectedID});
+		this.focusOnAButton(selectedID);
+	};
+
+	private handleDownArrow() {
+		let selectedID = this.state.mouseKeyOnButton + 1;
+		if (selectedID === 5) {
```
mttwc: These numbers are incorrect (off by 1)

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -1,21 +1,65 @@
...
+
+class ModeButtonSelectorClass extends ComponentBase<ModeButtonSelectorState, ClipperStateProp> {
+	constructor(props) {
+		super(props);
+		this.setState({
+			mouseKeyOnButton: 1
+		});
+	}
+
+	private handleKeydown = (event) => {
```
mttwc: We've deeply coupled arrow-key navigation to the buttons. We should abstract key movement into another class where it can be reused across other groupings in the future. (If you were to add arrow key navigation to a different set of buttons, how would you avoid duplicating code?)

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -1,21 +1,65 @@
...
+		let selectedID = this.state.mouseKeyOnButton + 1;
+		if (selectedID === 5) {
+			selectedID = 1;
+		}
+		this.setState({ mouseKeyOnButton: selectedID});
+		this.focusOnAButton(selectedID);
+	};
+
+	private focusOnAButton(selectedID) {
+		const buttons = document.querySelectorAll("a[aria-setsize]");
```
mttwc: Bug: If there are more than 1 button grouping in the Clipper, we can arrow navigate into them

src/scripts/clipperUI/components/previewViewer/previewComponentBase.tsx
```diff
@@ -1,27 +1,80 @@
...
 import {AnnotationInput} from "../annotationInput";
 
-import * as _ from "lodash";
-
 export abstract class PreviewComponentBase<TState, TProps extends ClipperStateProp>
 	extends ComponentBase<TState, TProps> {
 
 	private static textAreaListenerAttached = false;
 
+	render() {
```
mttwc: This file is hard to review since it contains large chunks of code that has been moved around, as well as changes in that code. Can you please organize the method orders back to the original order?

src/scripts/clipperUI/components/sectionPicker.tsx
```diff
@@ -1,18 +1,14 @@
 /// <reference path="../../../../node_modules/onenotepicker/target/oneNotePicker.d.ts"/>
 
 import {Constants} from "../../constants";
-import {Settings} from "../../settings";
```
mttwc, line 4: This file is hard to review since it contains large chunks of code that has been moved around, as well as changes in that code. Can you please organize the method orders back to the original order?

src/scripts/clipperUI/mainController.tsx
```diff
@@ -246,31 +247,32 @@ export class MainControllerClass extends ComponentBase<MainControllerState, Main
...
-				return <ErrorDialogPanel message={Localization.getLocalizedString("WebClipper.Error.OrphanedWebClipperDetected") } buttons={buttons} />;
+				return <ErrorDialogPanel
+					message={Localization.getLocalizedString("WebClipper.Error.OrphanedWebClipperDetected")} buttons={buttons}/>;
 			case PanelType.Loading:
-				return <LoadingPanel clipperState={this.props.clipperState} />;
+				return <LoadingPanel clipperState={this.props.clipperState}/>;
 			case PanelType.SignInNeeded:
 				return <SignInPanel clipperState={this.props.clipperState}
-					onSignInInvoked={this.props.onSignInInvoked}/>;
+														onSignInInvoked={this.props.onSignInInvoked}/>;
```
mttwc: Too much indentation

src/scripts/clipperUI/mainController.tsx
```diff
@@ -307,17 +309,18 @@ export class MainControllerClass extends ComponentBase<MainControllerState, Main
...
+				let panels: any[] = [<SuccessPanel clipperState={this.props.clipperState}/>];
 
 				if (!this.state.ratingsPanelAnimationState) {
 					this.state.ratingsPanelAnimationState = new SmartValue<AnimationState>(AnimationState.Out);
 				}
 
 				if (this.props.clipperState.showRatingsPrompt) {
-					panels.push(<RatingsPanel clipperState={this.props.clipperState} animationState={this.state.ratingsPanelAnimationState} />);
+					panels.push(<RatingsPanel clipperState={this.props.clipperState}
+																		animationState={this.state.ratingsPanelAnimationState}/>);
```
mttwc: Remove extra line

src/scripts/clipperUI/mainController.tsx
```diff
@@ -348,11 +351,11 @@ export class MainControllerClass extends ComponentBase<MainControllerState, Main
 					// so we will not show the footer. If it doesn't require signout, show the Footer
 					return undefined;
 				}
-				/* falls through */
+			/* falls through */
```
mttwc: Add indentation back for the 'falls through' comments

src/scripts/clipperUI/panels/regionSelectingPanel.tsx
```diff
@@ -16,16 +13,19 @@ class RegionSelectingPanelClass extends ComponentBase<{}, ClipperStateProp> {
 			<div id={Constants.Ids.regionInstructionsContainer}>
 				<div className="regionClipPadding">
 					<div className="messageLabelContainer">
+						<div
+							className={Constants.Classes.srOnly} aria-live="polite"
+							aria-relevant="text">{Localization.getLocalizedString("WebClipper.Label.DragAndRelease")}</div>
 						<span className="informationLabelFont messageLabel"
-							style={Localization.getFontFamilyAsStyle(Localization.FontFamily.Light)}>
+									style={Localization.getFontFamilyAsStyle(Localization.FontFamily.Light)}>
```
mttwc: Too much indentation

src/scripts/clipperUI/panels/regionSelectingPanel.tsx
```diff
@@ -16,16 +13,19 @@ class RegionSelectingPanelClass extends ComponentBase<{}, ClipperStateProp> {
...
 							{Localization.getLocalizedString("WebClipper.Label.DragAndRelease")}
 						</span>
 					</div>
-					<a id={Constants.Ids.regionClipCancelButton} {...this.enableInvoke(this.handleCancelButton, 70) }>
+					<a id={Constants.Ids.regionClipCancelButton} {...this.enableInvoke(this.handleCancelButton, 70)}>
 						<div className="wideButtonContainer">
 							<span className="wideButtonFont wideActionButton"
-								style={Localization.getFontFamilyAsStyle(Localization.FontFamily.Semibold)}>
-								{Localization.getLocalizedString("WebClipper.Action.BackToHome") }
+										style={Localization.getFontFamilyAsStyle(Localization.FontFamily.Semibold)}>
```
mttwc: Too much indentation

src/tests/assert.ts
```diff
@@ -25,6 +25,20 @@ export module Assert {
 		}
 	}
 
+	export function posInSetIsDecremental(elementIds: string[]) {
+		let elementsInExpectedPosInSetOrder = elementIds.map(id => {return{name: id, elem: document.getElementById(id) }; });
```
mttwc, line 29: You shouldn't need a return in the map here, you can do something like
id => {name: id, elem: document.getElementById(id)}

src/tests/clipperUI/components/modeButtonSelector_tests.tsx
```diff
@@ -161,14 +161,14 @@ export class ModeButtonSelectorTests extends TestModule {
 			strictEqual(buttonElements[3].id, TestConstants.Ids.bookmarkButton, "The fourth button should be the bookmark button");
 		});
 
-		test("The tabbing should flow in element order, assuming they are all available", () => {
+		test("The aria-posinset attribute should flow in element order, assuming they are all available", () => {
 			let startingState = MockProps.getMockClipperState();
```
mttwc, line 165: There should be other tests in this file that describe various groups where the buttons are different. Tests for those scenarios will need to be added as well. Right now your code will only work in the base-case where the user is clipping a normal web page, but what about pdf/etc?

src/tests/clipperUI/components/modeButton_tests.tsx
```diff
@@ -66,12 +66,20 @@ export class ModeButtonTests extends TestModule {
...
 			MithrilUtils.mountToFixture(this.defaultComponent);
 
 			let modeButton = MithrilUtils.getFixture().firstElementChild as HTMLElement;
-			strictEqual(modeButton.tabIndex, this.mockModeButtonProps.tabIndex,
-				"The mode button's tab index should be: " + this.mockModeButtonProps.tabIndex);
+			strictEqual(modeButton.getAttribute("aria-posinset"), this.mockModeButtonProps["aria-posinset"],
+				"The mode button's aria-pos attribute should be: " + this.mockModeButtonProps.tabIndex);
+		});
+
+		test("A button's aria-setsize attribute should be 4", () => {
```
mttwc: This should not always be the case, as mentioned in other comments

mttwc (changes requested):
Some overall thoughts I'd like to add on this PR:

- "Abstraction": when does it make sense for one component to know about another component? As a rule of thumb in the react world, a component should only know about its children (and not its siblings or its parents)
- There was an assumption made that there's 4 mode buttons in the Clipper. What if we were to change that number in the future? What if the number is determined at runtime instead of compile time? How would you write your code to be resilient to that?
- Arrow navigation is expected to be reused across the Clipper. How can you abstract it out so that it's reusable in different groups of components?

src/scripts/clipperUI/componentBase.ts
```diff
@@ -59,14 +59,14 @@ export abstract class ComponentBase<TState, TProps> {
...
 	/*
-	 * Helper which handles tabIndex, clicks, and keyboard navigation.
-	 *
-	 * Also hides the outline if they are using a mouse, but shows it if they are using the keyboard
-	 * (idea from http://www.paciellogroup.com/blog/2012/04/how-to-remove-css-outlines-in-an-accessible-manner/)
-	 *
-	 * Example use:
-	 *      <a id="myCoolButton" {...this.enableInvoke(this.myButtonHandler, 0)}>Click Me</a>
-	 */
+   * Helper which handles tabIndex, clicks, and keyboard navigation.
```
mttwc: Revert this block (we use tabs in our code, not spaces).

src/scripts/clipperUI/componentBase.ts
```diff
@@ -105,6 +105,25 @@ export abstract class ComponentBase<TState, TProps> {
 				} else if (e.which === Constants.KeyCodes.tab) {
 					// Since they are using the keyboard, revert to the default value of the outline so it is visible
 					element.style.outlineStyle = "";
+				} else if (e.which === Constants.KeyCodes.up) {
+					if (element.hasAttribute("data-ariaModeButtonSet")) {
```
mttwc: The ComponentBase is the superclass of all components in the Clipper's UI. By adding a reference to the modeButton here, we've broken abstraction. Additionally, if we wish to add directional-key navigation to other parts of the Clipper, this code is not generalized enough to be reusable. Think about how you can rework this code to not be coupled to modeButtons, while making it generalized/extensible enough to be consumed by other parts of the Clipper that might need it

src/scripts/clipperUI/componentBase.ts
```diff
@@ -105,6 +105,25 @@ export abstract class ComponentBase<TState, TProps> {
...
+							let nextPosInSet = parseInt(element.getAttribute("aria-posinset"), 10) - 1;
+							const buttons = document.querySelectorAll("a[data-ariaModeButtonSet]");
```
mttwc: The string data-ariaModeButtonSet is duplicated several times. This is prone to bugs as a dev might forget to change all of them if the attribute gets renamed.

src/scripts/clipperUI/componentBase.ts
```diff
@@ -105,6 +105,25 @@ export abstract class ComponentBase<TState, TProps> {
...
+							if (nextPosInSet === 0) {
+								nextPosInSet = buttons.length;
```
mttwc: Why set the next pos to buttons.length?

src/scripts/clipperUI/componentBase.ts
```diff
@@ -115,10 +134,24 @@ export abstract class ComponentBase<TState, TProps> {
...
+				selectable.style.outlineStyle = "";
+				let ariaIntForEach = parseInt(selectable.getAttribute("aria-posinset"), 10);
+				if (ariaIntForEach === nextPosInSet) {
+					selectable.focus();
+				}
+			}
+		}
+	}
+
+// Note: currently all components NEED either a child or attribute to work with the MSX transformer.
```
mttwc: Fix whitespace

src/scripts/clipperUI/componentBase.ts
```diff
@@ -115,10 +134,24 @@ export abstract class ComponentBase<TState, TProps> {
 		};
 	}
 
-	// Note: currently all components NEED either a child or attribute to work with the MSX transformer.
+	private static focusOnButton(buttons, nextPosInSet) {
```
mttwc: Missing parameter types

src/scripts/clipperUI/componentBase.ts
```diff
@@ -105,6 +105,25 @@ export abstract class ComponentBase<TState, TProps> {
...
+							const buttons = document.querySelectorAll("a[data-ariaModeButtonSet]");
+							if (nextPosInSet === 0) {
+								nextPosInSet = buttons.length;
+							}
+							ComponentBase.focusOnButton(buttons, nextPosInSet);
+					}
+
+				} else if (e.which === Constants.KeyCodes.down) {
+					if (element.hasAttribute("data-ariaModeButtonSet")) {
+							let nextPosInSet = parseInt(element.getAttribute("aria-posinset"), 10) + 1;
```
mttwc: Too much indentation

src/scripts/clipperUI/componentBase.ts
```diff
@@ -105,6 +105,25 @@ export abstract class ComponentBase<TState, TProps> {
...
+							const buttons = document.querySelectorAll("a[data-ariaModeButtonSet]");
+							if (nextPosInSet === (buttons.length + 1) ) {
```
mttwc: This can be hard to read since we've essentially added +1 to make nextPosInSet, then added +1 to the if statement. We can instead move the if-statement earlier before adding the +1, then checking if the currentPosInSet is the maximum allowed value.

mttwc: Same applies to the above block (up-key) where we subtract before checking if nextPosInSet === 0

src/scripts/clipperUI/componentBase.ts
```diff
@@ -115,10 +134,24 @@ export abstract class ComponentBase<TState, TProps> {
 		};
 	}
 
-	// Note: currently all components NEED either a child or attribute to work with the MSX transformer.
+	private static focusOnButton(buttons, nextPosInSet) {
+		if (buttons.length > 0) {
```
mttwc: This if statement is unnecessary. The for-loop will not execute if length === 0.

src/scripts/clipperUI/componentBase.ts
```diff
@@ -115,10 +134,24 @@ export abstract class ComponentBase<TState, TProps> {
...
+			for (let i = 0; i < buttons.length; i++) {
+				let selectable = buttons[i] as HTMLElement;
```
mttwc: You will not need the `as HTMLElement` if the parameters are typed in the method declaration

src/scripts/clipperUI/componentBase.ts
```diff
@@ -115,10 +134,24 @@ export abstract class ComponentBase<TState, TProps> {
...
+				selectable.style.outlineStyle = "";
```
mttwc: Why is this needed?

src/scripts/clipperUI/components/modeButton.tsx
```diff
@@ -23,18 +25,18 @@ class ModeButtonClass extends ComponentBase<{}, ModeButtonProps> {
...
-				{...this.enableInvoke(this.buttonHandler, this.props.tabIndex) }>
-				<img className="icon" src={this.props.imgSrc} />
-				<span className="label buttonLabelFont"
-					style={Localization.getFontFamilyAsStyle(Localization.FontFamily.Regular)}>
+			<a
+				className={className} role="option" data-ariaModeButtonSet={Constants.AriaSet.modeButtonSet}
+				id={idName} title={this.props.tooltipText ? this.props.tooltipText : ""} aria-setsize={this.props["aria-setsize"]} aria-posinset={this.props["aria-posinset"]}
+				{...this.enableInvoke(this.buttonHandler, this.props.tabIndex)} tabIndex={this.props.tabIndex}
+				aria-selected={this.props.selected}>
+			<img className="icon" src={this.props.imgSrc} />
```
mttwc: Fix indentation

src/scripts/clipperUI/components/modeButton.tsx
```diff
@@ -23,18 +25,18 @@ class ModeButtonClass extends ComponentBase<{}, ModeButtonProps> {
...
-			<a className={className} aria-selected={this.props.selected} role="button"
-				id={idName} title={this.props.tooltipText ? this.props.tooltipText : ""}
-				{...this.enableInvoke(this.buttonHandler, this.props.tabIndex) }>
-				<img className="icon" src={this.props.imgSrc} />
-				<span className="label buttonLabelFont"
-					style={Localization.getFontFamilyAsStyle(Localization.FontFamily.Regular)}>
+			<a
+				className={className} role="option" data-ariaModeButtonSet={Constants.AriaSet.modeButtonSet}
+				id={idName} title={this.props.tooltipText ? this.props.tooltipText : ""} aria-setsize={this.props["aria-setsize"]} aria-posinset={this.props["aria-posinset"]}
+				{...this.enableInvoke(this.buttonHandler, this.props.tabIndex)} tabIndex={this.props.tabIndex}
```
mttwc: tabIndex attribute is not needed. One job of `enableInvoke` is to set the tabIndex on the element.

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -35,12 +29,16 @@ class ModeButtonSelectorClass extends ComponentBase<{}, ClipperStateProp> {
...
+			imgSrc: ExtensionUtils.getImageResourceUrl("pdf.png"),
+			label: Localization.getLocalizedString("WebClipper.ClipType.Pdf.Button"),
+			myMode: ClipMode.Pdf,
+			selected: buttonSelected,
+			onModeSelected: this.onModeSelected.bind(this),
+			tooltipText: Localization.getLocalizedString("WebClipper.ClipType.Pdf.Button.Tooltip")
+		};
 	}
 
 	private getAugmentationModeButton(currentMode: ClipMode) {
```
mttwc: It's not being done yet, but we should add return types to each of these methods to add type safety to the mode button props being returned

mttwc: One option is to create a generic props interface (e.g., called AriaSetItemProps) that represents an item in an aria set, and have it have these two aria properties. Props that belong to directional-key navigatable components can then have their props extend that interface.

mttwc: The new ModeButtonProps type can then be a union of those two types:
e.g., ModeButtonProps = AriaSetItemProps & ModeButtonPropsWithoutAriaSetItemProps (or whatever name is better). This method can then return ModeButtonPropsWithoutAriaSetItemProps

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -51,25 +49,32 @@ class ModeButtonSelectorClass extends ComponentBase<{}, ClipperStateProp> {
...
 			return undefined;
 		}
-
-		return <ModeButton imgSrc={ExtensionUtils.getImageResourceUrl("fullpage.png")}
-			label={Localization.getLocalizedString("WebClipper.ClipType.ScreenShot.Button")}
-			myMode={ClipMode.FullPage} tabIndex={40}
-			selected={currentMode === ClipMode.FullPage}
-			onModeSelected={this.onModeSelected.bind(this) }
-			tooltipText={Localization.getLocalizedString("WebClipper.ClipType.ScreenShot.Button.Tooltip")}/>;
+		let buttonSelected: boolean = currentMode === ClipMode.FullPage;
```
mttwc: Was there a particular reason as to why buttonSelected was made into its own variable in this and other methods?

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -51,25 +49,32 @@ class ModeButtonSelectorClass extends ComponentBase<{}, ClipperStateProp> {
...
+			imgSrc: ExtensionUtils.getImageResourceUrl(augmentationType + ".png"),
+			label: augmentationLabel,
+			myMode: ClipMode.Augmentation,
+			selected: buttonSelected,
+			onModeSelected: this.onModeSelected.bind(this),
+			tooltipText: augmentationTooltip
+		};
 	}
 
 	private getFullPageModeButton(currentMode: ClipMode) {
```
mttwc: We should rename these methods. They no longer return buttons

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -95,37 +104,68 @@ class ModeButtonSelectorClass extends ComponentBase<{}, ClipperStateProp> {
...
+			onModeSelected: this.onModeSelected.bind(this),
+			tooltipText: Localization.getLocalizedString("WebClipper.ClipType.Bookmark.Button.Tooltip")
+		};
 	}
 
-	public render() {
+	private getListOfButtons() {
 		let currentMode = this.props.clipperState.currentMode.get();
 
+		let targets = [
```
mttwc: The variable name could be more descriptive. propsForButtons maybe?

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -95,37 +104,68 @@ class ModeButtonSelectorClass extends ComponentBase<{}, ClipperStateProp> {
...
+			this.getBookmarkModeButton(currentMode),
+			this.getPdfModeButton(currentMode),
+		];
+
+		let buttons = [
+			this.getScreenReaderOnlyElementThatAnnouncesCurrentMode(currentMode),
+		];
+
+		if (targets) {
+			let onPage = targets.filter(attributes => attributes !== undefined);
```
mttwc: The variable name could be more descriptive (onPage sounds more like a boolean)

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -95,37 +104,68 @@ class ModeButtonSelectorClass extends ComponentBase<{}, ClipperStateProp> {
...
+			for (let attributes of onPage) {
+				if (attributes === undefined) {
```
mttwc: We don't expect any of these to be undefined

src/scripts/clipperUI/components/modeButtonSelector.tsx
mttwc: The filter will not filter out items that are `null`. Use attributes => !!attributes

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -95,37 +104,68 @@ class ModeButtonSelectorClass extends ComponentBase<{}, ClipperStateProp> {
...
+					return undefined;
+				}
+				let ariaPos = onPage.indexOf(attributes) + 1;
```
mttwc: indexOf is not efficient (it's O(n)). You can get the index for free if you use a normal for loop.

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -95,37 +104,68 @@ class ModeButtonSelectorClass extends ComponentBase<{}, ClipperStateProp> {
...
+
+		let buttons = [
+			this.getScreenReaderOnlyElementThatAnnouncesCurrentMode(currentMode),
+		];
+
+		if (targets) {
+			let onPage = targets.filter(attributes => attributes !== undefined);
+			for (let attributes of onPage) {
+				if (attributes === undefined) {
+					return undefined;
```
mttwc: Why return undefined?

src/scripts/clipperUI/components/modeButton.tsx
```diff
@@ -1,14 +1,16 @@
...
+import {ComponentBase} from "../componentBase";
 
 export interface ModeButtonProps {
 	imgSrc: string;
 	label: string;
 	myMode: ClipMode;
 	tabIndex: number;
 	selected?: boolean;
+	"aria-posinset": string;
+	"aria-setsize": string;
```
mttwc: Why are these two props strings?

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -95,37 +104,68 @@ class ModeButtonSelectorClass extends ComponentBase<{}, ClipperStateProp> {
...
+				}
+				let ariaPos = onPage.indexOf(attributes) + 1;
+				buttons.push( <ModeButton imgSrc={attributes.imgSrc} label={attributes.label} myMode={attributes.myMode} selected={attributes.selected} onModeSelected={attributes.onModeSelected} tooltipText={attributes.tooltipText} aria-setsize={onPage.length.toString()} aria-posinset={ariaPos.toString()} tabIndex={attributes.selected ? 40 : "" } /> );
```
mttwc: We can shorten this line, it's pretty long

src/scripts/clipperUI/mainController.tsx
```diff
@@ -246,31 +238,31 @@ export class MainControllerClass extends ComponentBase<MainControllerState, Main
...
 					if (this.props.clipperState.pdfPreviewInfo.shouldDistributePages) {
-						return <ClippingPanelWithProgressIndicator clipperState= { this.props.clipperState } />;
+						return <ClippingPanelWithProgressIndicator clipperState={this.props.clipperState}/>;
 					}
 
 					return <ClippingPanelWithDelayedMessage
-						clipperState={this.props.clipperState}
-						delay={Constants.Settings.pdfClippingMessageDelay}
-						message={Localization.getLocalizedString("WebClipper.ClipType.Pdf.ProgressLabelDelay")} />;
+									clipperState={this.props.clipperState}
```
mttwc: Too much indentation in some places in this file (including this one)

mttwc (changes requested):

src/scripts/clipperUI/componentBase.ts
JaneHazen: To reset the outline style if it was clicked on previously (the mousedown function sets the outline style to "none")

src/scripts/clipperUI/componentBase.ts
JaneHazen: so the user can scroll through the options in a wheel instead of just once-- this is maybe more design preference? is there a standard either way?

src/scripts/clipperUI/componentBase.ts
```diff
@@ -59,15 +59,15 @@ export abstract class ComponentBase<TState, TProps> {
...
-	 * Helper which handles tabIndex, clicks, and keyboard navigation.
-	 *
-	 * Also hides the outline if they are using a mouse, but shows it if they are using the keyboard
-	 * (idea from http://www.paciellogroup.com/blog/2012/04/how-to-remove-css-outlines-in-an-accessible-manner/)
-	 *
-	 * Example use:
-	 *      <a id="myCoolButton" {...this.enableInvoke(this.myButtonHandler, 0)}>Click Me</a>
-	 */
-	public enableInvoke(handleMethod: Function, tabIndex = 0, args?: any, idOverride: string = undefined) {
+   	* Helper which handles tabIndex, clicks, and keyboard navigation.
```
mttwc: This comment block is styled incorrectly (needs to use tabs everywhere, not spaces). Each asterisk between /* and */ should have an extra space like before.

src/scripts/clipperUI/componentBase.ts
```diff
@@ -59,15 +59,15 @@ export abstract class ComponentBase<TState, TProps> {
...
+	*
+	* Also hides the outline if they are using a mouse, but shows it if they are using the keyboard
+	* (idea from http://www.paciellogroup.com/blog/2012/04/how-to-remove-css-outlines-in-an-accessible-manner/)
+	*
+	* Example use:
+	*      <a id="myCoolButton" {...this.enableInvoke(this.myButtonHandler, 0)}>Click Me</a>
+	*/
+	public enableInvoke(handleMethod: Function, tabIndex = 0 , args?: any, idOverride: string = undefined, ariaSet: string = undefined) {
```
mttwc: Remove extra space between 0 and , after tabIndex

src/scripts/clipperUI/componentBase.ts
```diff
@@ -105,6 +105,26 @@ export abstract class ComponentBase<TState, TProps> {
 				} else if (e.which === Constants.KeyCodes.tab) {
 					// Since they are using the keyboard, revert to the default value of the outline so it is visible
 					element.style.outlineStyle = "";
+				} else if (e.which === Constants.KeyCodes.up) {
+					if (element.hasAttribute("data-" + ariaSet)) {
+							let nextPosInSet = parseInt(element.getAttribute("aria-posinset"), 10) - 1;
```
mttwc: Too much indentation

src/scripts/clipperUI/componentBase.ts
```diff
@@ -105,6 +105,26 @@ export abstract class ComponentBase<TState, TProps> {
 				} else if (e.which === Constants.KeyCodes.tab) {
 					// Since they are using the keyboard, revert to the default value of the outline so it is visible
 					element.style.outlineStyle = "";
+				} else if (e.which === Constants.KeyCodes.up) {
+					if (element.hasAttribute("data-" + ariaSet)) {
```
mttwc: Rather than setting an attribute whose key uniquely identifies the set, why not have a predictable attribute name and assign it a value of the set name?

src/scripts/clipperUI/componentBase.ts
```diff
@@ -105,6 +105,26 @@ export abstract class ComponentBase<TState, TProps> {
...
+							let nextPosInSet = parseInt(element.getAttribute("aria-posinset"), 10) - 1;
+							const buttons = document.querySelectorAll("a[data-" + ariaSet + "]");
+							if (nextPosInSet === 0) {
```
mttwc: Please see comments around this entire block of code. I don't think the suggestions were applied. E.g., by assigning nextPosInSet to buttons.length, wouldn't you loop back around to the last item in the set? We wish to do nothing in this case.

src/scripts/clipperUI/componentBase.ts
```diff
@@ -115,10 +135,22 @@ export abstract class ComponentBase<TState, TProps> {
 		};
 	}
 
+	private static focusOnButton(buttons: NodeListOf<Element>, nextPosInSet: number) {
+		for (let i = 0; i < buttons.length; i++) {
+			let selectable = buttons[i] as HTMLElement;
+			selectable.style.outlineStyle = "";
```
mttwc: Please see comments in previous review around this block

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -1,28 +1,39 @@
...
+	imgSrc: string;
+	label: string;
+	myMode: ClipMode;
+	selected?: boolean;
+	tabIndex?: number;
+	onModeSelected: (modeButton: ClipMode) => void;
+	tooltipText?: string;
+}
+
+export interface AriaPropsForModeButton {
```
mttwc: This should belong in another file, since it has nothing to do with modebuttons anymore (abstraction)

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -1,28 +1,39 @@
...
 class ModeButtonSelectorClass extends ComponentBase<{}, ClipperStateProp> {
+
 	onModeSelected(newMode: ClipMode) {
 		this.props.clipperState.setState({
 			currentMode: this.props.clipperState.currentMode.set(newMode)
 		});
 	};
 
-	private getScreenReaderOnlyElementThatAnnouncesCurrentMode(currentMode: ClipMode) {
+	private getScreenReaderOnlyElementPropsThatAnnouncesCurrentMode(currentMode: ClipMode) {
```
mttwc: Standardize naming of these methods (in some of these, props is the last word, but not always)

src/scripts/clipperUI/components/modeButtonSelector.tsx
mttwc: Also the name should not reference ModeButton at all, since it's generalized enough to be reused

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -1,28 +1,39 @@
...
-
 import {Localization} from "../../localization/localization";
-
 import {ClipMode} from "../clipMode";
 import {ClipperStateProp} from "../clipperState";
 import {ComponentBase} from "../componentBase";
-
 import {ModeButton} from "./modeButton";
 
+export interface NoAriaPropsForModeButton {
```
mttwc: ModeButtonSelector should not have ModeButtonProps and similar interfaces. These should be declared in modeButton.tsx

mttwc: Also I wonder if there's a better name for this interface ... it could be clearer

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -31,101 +42,150 @@ class ModeButtonSelectorClass extends ComponentBase<{}, ClipperStateProp> {
...
+				visibleButtons.push(<ModeButton imgSrc={attributes.imgSrc} label={attributes.label} myMode={attributes.myMode}
+								selected={attributes.selected} onModeSelected={attributes.onModeSelected}
+								tooltipText={attributes.tooltipText} aria-setsize={propsForVisibleButtons.length}
+								aria-posinset={ariaPos} tabIndex={attributes.selected ? 40 : undefined} />);
+			}
+			return visibleButtons;
+		}
+	}
+
+	private static getIndexOf(list, element): number {
```
mttwc: This is still O(n). Rather than using for ... of in the above block, using a simple for loop will remove the need to iterate through the list again to find the index.

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -31,101 +42,150 @@ class ModeButtonSelectorClass extends ComponentBase<{}, ClipperStateProp> {
...
+		for (let i = 0; i < list.length; i++) {
+			if (list[i] === element) {
+				return i;
+			}
+		}
+	}
+
+	public render() {
+
```
mttwc: Remove extra newline

mttwc (changes requested):

src/scripts/clipperUI/componentBase.ts
```diff
@@ -105,6 +104,20 @@ export abstract class ComponentBase<TState, TProps> {
 				} else if (e.which === Constants.KeyCodes.tab) {
 					// Since they are using the keyboard, revert to the default value of the outline so it is visible
 					element.style.outlineStyle = "";
+				} else if (e.which === Constants.KeyCodes.up) {
+					if (element.hasAttribute("data-ariaSet")) {
+							let nextPosInSet = parseInt(element.getAttribute("aria-posinset"), 10) - 1;
```
mttwc: Too much indentation

src/scripts/clipperUI/componentBase.ts
```diff
@@ -67,7 +66,7 @@ export abstract class ComponentBase<TState, TProps> {
 	 * Example use:
 	 *      <a id="myCoolButton" {...this.enableInvoke(this.myButtonHandler, 0)}>Click Me</a>
 	 */
-	public enableInvoke(handleMethod: Function, tabIndex = 0, args?: any, idOverride: string = undefined) {
+	public enableInvoke(handleMethod: Function, tabIndex = 0, args?: any, idOverride: string = undefined, ariaSet: string = undefined) {
```
mttwc: Rename ariaSet to something that indicates that hints at directional key nav? ariaSet is not descriptive of what it's used for

src/scripts/clipperUI/componentBase.ts
```diff
@@ -105,6 +104,20 @@ export abstract class ComponentBase<TState, TProps> {
 				} else if (e.which === Constants.KeyCodes.tab) {
 					// Since they are using the keyboard, revert to the default value of the outline so it is visible
 					element.style.outlineStyle = "";
+				} else if (e.which === Constants.KeyCodes.up) {
+					if (element.hasAttribute("data-ariaSet")) {
+							let nextPosInSet = parseInt(element.getAttribute("aria-posinset"), 10) - 1;
+							const buttons = document.querySelectorAll("a[data-ariaSet=" + ariaSet.toString() + "]");
```
mttwc: Why is toString needed?

mttwc: Also this will throw an exception if ariaSet is undefined (ariaSet is an optional parameter)

src/scripts/clipperUI/componentBase.ts
```diff
@@ -105,6 +104,20 @@ export abstract class ComponentBase<TState, TProps> {
...
+							ComponentBase.focusOnButton(buttons, nextPosInSet);
+					}
+
+				} else if (e.which === Constants.KeyCodes.down) {
+					if (element.hasAttribute("data-ariaSet")) {
+						let posInSet = parseInt(element.getAttribute("aria-posinset"), 10);
+						let nextPosInSet = posInSet + 1;
+						const buttons = document.querySelectorAll("a[data-ariaSet=" + ariaSet.toString() + "]");
```
mttwc: We can move the querySelectorAll call to the focusOnButton method (i.e., focusOnButton's 1st param would be the name of the set, rather than the list of elements)

src/scripts/clipperUI/componentBase.ts
```diff
@@ -115,10 +128,22 @@ export abstract class ComponentBase<TState, TProps> {
 		};
 	}
 
+	private static focusOnButton(buttons: NodeListOf<Element>, nextPosInSet: number) {
+		for (let i = 0; i < buttons.length; i++) {
+			let selectable = buttons[i] as HTMLElement;
+			selectable.style.outlineStyle = "";
```
mttwc: We should probably only do this to the winning element

src/scripts/clipperUI/componentBase.ts
```diff
@@ -115,10 +128,22 @@ export abstract class ComponentBase<TState, TProps> {
...
+			let ariaIntForEach = parseInt(selectable.getAttribute("aria-posinset"), 10);
+			if (ariaIntForEach === nextPosInSet) {
+				selectable.focus();
```
mttwc: We can return here

src/scripts/clipperUI/components/modeButton.tsx
```diff
@@ -1,19 +1,22 @@
...
 	myMode: ClipMode;
-	tabIndex: number;
 	selected?: boolean;
+	tabIndex?: number;
 	onModeSelected: (modeButton: ClipMode) => void;
 	tooltipText?: string;
 }
 
-class ModeButtonClass extends ComponentBase<{}, ModeButtonProps> {
+export interface PropsForModeButton extends PropsForModeElementNoAriaGrouping, PropsForAriaGrouping { }
```
mttwc: Rename this to ModeButtonsProps to match the naming convention we use for all props

src/scripts/clipperUI/components/modeButton.tsx
```diff
@@ -23,18 +26,17 @@ class ModeButtonClass extends ComponentBase<{}, ModeButtonProps> {
...
-				id={idName} title={this.props.tooltipText ? this.props.tooltipText : ""}
-				{...this.enableInvoke(this.buttonHandler, this.props.tabIndex) }>
-				<img className="icon" src={this.props.imgSrc} />
-				<span className="label buttonLabelFont"
-					style={Localization.getFontFamilyAsStyle(Localization.FontFamily.Regular)}>
+			<a className={className} role="option" data-ariaSet={Constants.AriaSet.modeButtonSet}
+				id={idName} title={this.props.tooltipText ? this.props.tooltipText : ""} aria-setsize={this.props["aria-setsize"]} aria-posinset={this.props["aria-posinset"]}
+				{...this.enableInvoke(this.buttonHandler, this.props.tabIndex, undefined, undefined,  Constants.AriaSet.modeButtonSet)}
+				aria-selected={this.props.selected}>
+			<img className="icon" src={this.props.imgSrc} />
```
mttwc: Fix indentation

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -31,101 +26,142 @@ class ModeButtonSelectorClass extends ComponentBase<{}, ClipperStateProp> {
...
+			label: Localization.getLocalizedString(this.getRegionButtonPropsLabel()),
+			myMode: ClipMode.Region,
+			selected: currentMode === ClipMode.Region,
+			onModeSelected: this.onModeSelected.bind(this),
+			tooltipText: Localization.getLocalizedString("WebClipper.ClipType.MultipleRegions.Button.Tooltip")
+		};
 	}
 
-	private getRegionModeButtonLabel(): string {
+	private getRegionButtonPropsLabel(): string {
```
mttwc: This should not be renamed (it has nothing to do with props)

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -31,101 +26,142 @@ class ModeButtonSelectorClass extends ComponentBase<{}, ClipperStateProp> {
...
+			this.getSelectionButtonProps(currentMode),
+			this.getBookmarkButtonProps(currentMode),
+			this.getPdfButtonProps(currentMode),
+		];
+
+		let visibleButtons = [
+			this.getScreenReaderOnlyElementPropsThatAnnouncesCurrentMode(currentMode),
+		];
+
+		if (buttonProps) {
```
mttwc: This if statement isn't needed

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -31,101 +26,142 @@ class ModeButtonSelectorClass extends ComponentBase<{}, ClipperStateProp> {
...
+			label: Localization.getLocalizedString("WebClipper.ClipType.Bookmark.Button"),
+			myMode: ClipMode.Bookmark,
+			selected: currentMode === ClipMode.Bookmark,
+			onModeSelected: this.onModeSelected.bind(this),
+			tooltipText: Localization.getLocalizedString("WebClipper.ClipType.Bookmark.Button.Tooltip")
+		};
 	}
 
-	public render() {
+	private getListOfButtons() {
```
mttwc: Add return type

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -31,101 +26,142 @@ class ModeButtonSelectorClass extends ComponentBase<{}, ClipperStateProp> {
...
+			for (let i = 0; i < propsForVisibleButtons.length; i++) {
+				let attributes = propsForVisibleButtons[i];
+				let ariaPos = i + 1;
+				visibleButtons.push(<ModeButton imgSrc={attributes.imgSrc} label={attributes.label} myMode={attributes.myMode}
+								selected={attributes.selected} onModeSelected={attributes.onModeSelected}
+								tooltipText={attributes.tooltipText} aria-setsize={propsForVisibleButtons.length}
+								aria-posinset={ariaPos} tabIndex={attributes.selected ? 40 : undefined} />);
+			}
+			return visibleButtons;
+		}
```
mttwc: It's possible that this method does not return a value

mttwc (changes requested):

src/scripts/clipperUI/components/modeButtonSelector.tsx
mttwc: (see above comment to add return type to the method, that will add stricter compile time checking which would prevent this issue)

src/scripts/clipperUI/componentBase.ts
```diff
@@ -67,7 +66,7 @@ export abstract class ComponentBase<TState, TProps> {
 	 * Example use:
 	 *      <a id="myCoolButton" {...this.enableInvoke(this.myButtonHandler, 0)}>Click Me</a>
 	 */
-	public enableInvoke(handleMethod: Function, tabIndex = 0, args?: any, idOverride: string = undefined) {
+	public enableInvoke(handleMethod: Function, tabIndex = 0, args?: any, idOverride: string = undefined, ariaForArrowKeyNav: string = undefined) {
```
mttwc: Call it setNameForArrowKeyNav? aria is a name of technical specification

src/scripts/clipperUI/componentBase.ts
```diff
@@ -105,6 +104,18 @@ export abstract class ComponentBase<TState, TProps> {
 				} else if (e.which === Constants.KeyCodes.tab) {
 					// Since they are using the keyboard, revert to the default value of the outline so it is visible
 					element.style.outlineStyle = "";
+				} else if (e.which === Constants.KeyCodes.up) {
+					if (element.hasAttribute("data-ariaForArrowKeyNav")) {
+						let nextPosInSet = parseInt(element.getAttribute("aria-posinset"), 10) - 1;
```
mttwc: We should do nothing if current posinset is 1

src/scripts/clipperUI/componentBase.ts
```diff
@@ -105,6 +104,18 @@ export abstract class ComponentBase<TState, TProps> {
...
+						ComponentBase.focusOnButton(ariaForArrowKeyNav, nextPosInSet);
+					}
+
+				} else if (e.which === Constants.KeyCodes.down) {
+					if (element.hasAttribute("data-ariaForArrowKeyNav")) {
+						let posInSet = parseInt(element.getAttribute("aria-posinset"), 10);
```
mttwc: We should do nothing if current posinset is the setsize

src/scripts/clipperUI/componentBase.ts
```diff
@@ -115,10 +126,23 @@ export abstract class ComponentBase<TState, TProps> {
 		};
 	}
 
+	private static focusOnButton(ariaForArrowKeyNav: string, nextPosInSet: number) {
+		const buttons = document.querySelectorAll("a[data-ariaForArrowKeyNav=" + ariaForArrowKeyNav + "]");
```
mttwc: The attribute name should probably be a constant somewhere, so that if we change it, we don't have to change the name everywhere in all the components

src/scripts/clipperUI/componentBase.ts
```diff
@@ -115,10 +126,23 @@ export abstract class ComponentBase<TState, TProps> {
...
+		for (let i = 0; i < buttons.length; i++) {
+			let selectable = buttons[i] as HTMLElement;
+			let ariaIntForEach = parseInt(selectable.getAttribute("aria-posinset"), 10);
+			if (ariaIntForEach === nextPosInSet) {
+				selectable.style.outlineStyle = "";
+				selectable.focus();
```
mttwc: Return here. We don't have to iterate through the entire list most of the time.

src/scripts/clipperUI/components/modeButton.tsx
```diff
@@ -23,18 +26,19 @@ class ModeButtonClass extends ComponentBase<{}, ModeButtonProps> {
...
 				id={idName} title={this.props.tooltipText ? this.props.tooltipText : ""}
-				{...this.enableInvoke(this.buttonHandler, this.props.tabIndex) }>
-				<img className="icon" src={this.props.imgSrc} />
+				aria-setsize={this.props["aria-setsize"]} aria-posinset={this.props["aria-posinset"]}
+				{...this.enableInvoke(this.buttonHandler, this.props.tabIndex, undefined, undefined, Constants.AriaSet.modeButtonSet)}
+				aria-selected={this.props.selected}>
+				<img className="icon" src={this.props.imgSrc}/>
 				<span className="label buttonLabelFont"
-					style={Localization.getFontFamilyAsStyle(Localization.FontFamily.Regular)}>
+							style={Localization.getFontFamilyAsStyle(Localization.FontFamily.Regular)}>
```
mttwc: Too much indentation

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -1,28 +1,22 @@
...
-
 import {ClipMode} from "../clipMode";
 import {ClipperStateProp} from "../clipperState";
 import {ComponentBase} from "../componentBase";
-
-import {ModeButton} from "./modeButton";
+import {ModeButton, PropsForModeElementNoAriaGrouping} from "./modeButton";
 
 class ModeButtonSelectorClass extends ComponentBase<{}, ClipperStateProp> {
+
```
mttwc: Remove extra whitespace

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -31,101 +25,140 @@ class ModeButtonSelectorClass extends ComponentBase<{}, ClipperStateProp> {
...
+			selected: currentMode === ClipMode.Selection,
+			onModeSelected: this.onModeSelected.bind(this),
+			tooltipText: Localization.getLocalizedString("WebClipper.ClipType.Selection.Button.Tooltip")
+		};
 	}
 
-	private getBookmarkModeButton(currentMode: ClipMode) {
-		if (this.props.clipperState.pageInfo.rawUrl.indexOf("file:///") === 0) {
+	private getBookmarkButtonProps(currentMode: ClipMode): PropsForModeElementNoAriaGrouping {
+		if (this.props.clipperState.pageInfo.rawUrl.slice(0, 8) === "file:///") {
```
mttwc: On hindsight, I still think indexOf is better simply because it's safer to use (despite probably being less efficient). Right now if the slice length and the string length and mismatched, if will always return false. It's an easy bug to introduce.

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -31,101 +25,140 @@ class ModeButtonSelectorClass extends ComponentBase<{}, ClipperStateProp> {
...
+		let visibleButtons = [
+			this.getScreenReaderThatAnnouncesCurrentModeProps(currentMode),
+		];
+
+		let propsForVisibleButtons = buttonProps.filter(attributes => !!attributes);
+		for (let i = 0; i < propsForVisibleButtons.length; i++) {
+			let attributes = propsForVisibleButtons[i];
+			let ariaPos = i + 1;
+			visibleButtons.push(<ModeButton imgSrc={attributes.imgSrc} label={attributes.label} myMode={attributes.myMode}
+																			selected={attributes.selected} onModeSelected={attributes.onModeSelected}
```
mttwc: Remove all these extra lines

src/scripts/clipperUI/mainController.tsx
```diff
@@ -246,31 +238,31 @@ export class MainControllerClass extends ComponentBase<MainControllerState, Main
...
 				});
-				return <ErrorDialogPanel message={Localization.getLocalizedString("WebClipper.Error.OrphanedWebClipperDetected") } buttons={buttons} />;
+				return <ErrorDialogPanel message={Localization.getLocalizedString("WebClipper.Error.OrphanedWebClipperDetected")} buttons={buttons}/>;
 			case PanelType.Loading:
-				return <LoadingPanel clipperState={this.props.clipperState} />;
+				return <LoadingPanel clipperState={this.props.clipperState}/>;
 			case PanelType.SignInNeeded:
 				return <SignInPanel clipperState={this.props.clipperState}
-					onSignInInvoked={this.props.onSignInInvoked}/>;
+														onSignInInvoked={this.props.onSignInInvoked}/>;
```
mttwc: Too much indentation

src/scripts/clipperUI/components/ariaGroupingProps.ts
```diff
@@ -0,0 +1,4 @@
+export interface PropsForAriaGrouping {
```
mttwc: Call this ariaSetProps.ts? Sets in aria have a very defined connotation so we shouldn't use "set" and "group" interchangeably

mttwc (changes requested):

src/scripts/clipperUI/componentBase.ts
```diff
@@ -67,7 +67,7 @@ export abstract class ComponentBase<TState, TProps> {
 	 * Example use:
 	 *      <a id="myCoolButton" {...this.enableInvoke(this.myButtonHandler, 0)}>Click Me</a>
 	 */
-	public enableInvoke(handleMethod: Function, tabIndex = 0, args?: any, idOverride: string = undefined) {
+	public enableInvoke(handleMethod: Function, tabIndex = 0, args?: any, idOverride: string = undefined, valueOfSetNameForArrowKeyNav: string = undefined) {
```
mttwc: 'valueOf' is redundant, a name like "setNameForArrowKeyNav" is fine

src/scripts/clipperUI/componentBase.ts
```diff
@@ -105,6 +105,27 @@ export abstract class ComponentBase<TState, TProps> {
 				} else if (e.which === Constants.KeyCodes.tab) {
 					// Since they are using the keyboard, revert to the default value of the outline so it is visible
 					element.style.outlineStyle = "";
+				} else if (e.which === Constants.KeyCodes.up) {
+					if (element.hasAttribute("data-" + Constants.AriaSet.setNameForArrowKeyNav)) {
+						let posInSet = parseInt(element.getAttribute("aria-posinset"), 10);
+						if (posInSet === 1) {
+							return;
+						}
+						const buttons = document.querySelectorAll("a[data-" + Constants.AriaSet.setNameForArrowKeyNav + "=" + valueOfSetNameForArrowKeyNav + "]");
```
mttwc: This code only works if the 'buttons' are html 'a' tags. We should allow this code to work for all elements.

src/scripts/clipperUI/componentBase.ts
mttwc: Note that enableInvoke returns a list of additional attributes to add to the element it is called on e.g., onclick, tabIndex. We can use the value of your new param to add the setNameForArrowKeyNav attribute to the element automatically, so elements don't have to add it manually as well (which is duplicate work)

src/scripts/clipperUI/componentBase.ts
```diff
@@ -105,6 +105,27 @@ export abstract class ComponentBase<TState, TProps> {
 				} else if (e.which === Constants.KeyCodes.tab) {
 					// Since they are using the keyboard, revert to the default value of the outline so it is visible
 					element.style.outlineStyle = "";
+				} else if (e.which === Constants.KeyCodes.up) {
```
mttwc: The direction key handlers should not execute if the new param is undefined (remember to add explicit checks for undefined params)

src/scripts/clipperUI/componentBase.ts
```diff
@@ -105,6 +105,27 @@ export abstract class ComponentBase<TState, TProps> {
...
+						let nextPosInSet = posInSet - 1;
+						ComponentBase.focusOnButton(buttons, nextPosInSet);
+					}
+
+				} else if (e.which === Constants.KeyCodes.down) {
+					if (element.hasAttribute("data-" + Constants.AriaSet.setNameForArrowKeyNav)) {
+						let posInSet = parseInt(element.getAttribute("aria-posinset"), 10);
+						let nextPosInSet = posInSet + 1;
+						const buttons = document.querySelectorAll("a[data-" + Constants.AriaSet.setNameForArrowKeyNav + "=" + valueOfSetNameForArrowKeyNav + "]");
+						if (posInSet === buttons.length) {
```
mttwc: Check the posInSet as soon as you get the value

src/scripts/clipperUI/componentBase.ts
```diff
@@ -105,6 +105,27 @@ export abstract class ComponentBase<TState, TProps> {
...
+						const buttons = document.querySelectorAll("a[data-" + Constants.AriaSet.setNameForArrowKeyNav + "=" + valueOfSetNameForArrowKeyNav + "]");
+						let nextPosInSet = posInSet - 1;
+						ComponentBase.focusOnButton(buttons, nextPosInSet);
+					}
+
+				} else if (e.which === Constants.KeyCodes.down) {
+					if (element.hasAttribute("data-" + Constants.AriaSet.setNameForArrowKeyNav)) {
+						let posInSet = parseInt(element.getAttribute("aria-posinset"), 10);
+						let nextPosInSet = posInSet + 1;
+						const buttons = document.querySelectorAll("a[data-" + Constants.AriaSet.setNameForArrowKeyNav + "=" + valueOfSetNameForArrowKeyNav + "]");
```
mttwc: See above (generalize this to work for all html element types)

src/scripts/clipperUI/components/ariaSetProps.ts
```diff
@@ -0,0 +1,4 @@
+export interface ariaSetProps {
```
mttwc: Name should be capitalized i.e., AriaSetProps, to follow the convention we use

src/scripts/clipperUI/components/modeButton.tsx
```diff
@@ -23,18 +26,18 @@ class ModeButtonClass extends ComponentBase<{}, ModeButtonProps> {
...
 			className += " selected";
 		}
-
 		let clipMode: string = ClipMode[this.props.myMode];
 		clipMode = clipMode[0].toLowerCase() + clipMode.slice(1);
 		let idName: string = clipMode + "Button";
 
 		return (
-			<a className={className} aria-selected={this.props.selected} role="button"
+			<a className={className} role="option" data-setnameforarrowkeynav={Constants.AriaSet.modeButtonSet}
```
mttwc: See my previous comment on adding this attribute using enableInvoke. This way you don't have to set it explicitly again (which also is resilient against attribute renaming, which this is susceptible to)

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -31,101 +24,138 @@ class ModeButtonSelectorClass extends ComponentBase<{}, ClipperStateProp> {
...
+		let visibleButtons = [
+			this.getScreenReaderThatAnnouncesCurrentModeProps(currentMode),
+		];
+
+		let propsForVisibleButtons = buttonProps.filter(attributes => !!attributes);
+		for (let i = 0; i < propsForVisibleButtons.length; i++) {
+			let attributes = propsForVisibleButtons[i];
+			let ariaPos = i + 1;
+			visibleButtons.push(<ModeButton imgSrc={attributes.imgSrc} label={attributes.label} myMode={attributes.myMode} selected={attributes.selected}
+									onModeSelected={attributes.onModeSelected} tooltipText={attributes.tooltipText} aria-setsize={propsForVisibleButtons.length} aria-posinset={ariaPos} tabIndex={attributes.selected ? 40 : undefined}/>);
```
mttwc: Too much indentation

src/scripts/clipperUI/components/modeButtonSelector.tsx
```diff
@@ -31,101 +24,138 @@ class ModeButtonSelectorClass extends ComponentBase<{}, ClipperStateProp> {
...
+
+		let visibleButtons = [
+			this.getScreenReaderThatAnnouncesCurrentModeProps(currentMode),
+		];
+
+		let propsForVisibleButtons = buttonProps.filter(attributes => !!attributes);
+		for (let i = 0; i < propsForVisibleButtons.length; i++) {
+			let attributes = propsForVisibleButtons[i];
+			let ariaPos = i + 1;
+			visibleButtons.push(<ModeButton imgSrc={attributes.imgSrc} label={attributes.label} myMode={attributes.myMode} selected={attributes.selected}
```
mttwc: I *think* you can use this syntax:
<ModeButton {...attributes} aria-setsize={propsForVisibleButtons.length} aria-posinset={ariaPos} tabIndex={attributes.selected ? 40 : undefined} />

mttwc (changes requested):

src/scripts/clipperUI/componentBase.ts
```diff
@@ -105,20 +105,61 @@ export abstract class ComponentBase<TState, TProps> {
 				} else if (e.which === Constants.KeyCodes.tab) {
 					// Since they are using the keyboard, revert to the default value of the outline so it is visible
 					element.style.outlineStyle = "";
+				} else if (e.which === Constants.KeyCodes.up) {
+					if ( setNameForArrowKeyNav === undefined) {
```
mttwc: We should avoid repeated code as much as possible. You can do that by adding the check before checking if e.which === up/down. Also, you should check for null and empty string cases, which can be done with `if (!setNameForArrowKeyNav)

src/scripts/clipperUI/components/modeButton.tsx
```diff
@@ -23,18 +26,18 @@ class ModeButtonClass extends ComponentBase<{}, ModeButtonProps> {
 		if (this.props.selected) {
 			className += " selected";
 		}
-
 		let clipMode: string = ClipMode[this.props.myMode];
 		clipMode = clipMode[0].toLowerCase() + clipMode.slice(1);
 		let idName: string = clipMode + "Button";
 
 		return (
-			<a className={className} aria-selected={this.props.selected} role="button"
```
mttwc, line 32: Did you intend to remove aria-selected?

mttwc (changes requested):

src/scripts/clipperUI/componentBase.ts
mttwc: Also please remove whitespace after `if ( `

src/scripts/clipperUI/componentBase.ts
```diff
@@ -105,25 +105,19 @@ export abstract class ComponentBase<TState, TProps> {
...
 					element.style.outlineStyle = "";
-				} else if (e.which === Constants.KeyCodes.up) {
-					if ( setNameForArrowKeyNav === undefined) {
-						return;
-					}
-					if (element.hasAttribute("data-" + Constants.AriaSet.setNameForArrowKeyNav)) {
-						let posInSet = parseInt(element.getAttribute("aria-posinset"), 10);
+				}
+
+				if (setNameForArrowKeyNav === undefined || setNameForArrowKeyNav === "") {
```
mttwc: Use if (!setNameForArrowKeyNav), which covers undefined, null, and "" cases

mttwc (approved):

PR #398, opened 2018-05-26
Alter keyboard navigation for accessibility

JaneHazen:
change enter to default action, space bar to selection

### DIFF
--- a/src/scripts/clipperUI/componentBase.ts
+++ b/src/scripts/clipperUI/componentBase.ts
@@ -67,7 +67,7 @@ export abstract class ComponentBase<TState, TProps> {
 	 * Example use:
 	 *      <a id="myCoolButton" {...this.enableInvoke(this.myButtonHandler, 0)}>Click Me</a>
 	 */
-	public enableInvoke(handleMethod: Function, tabIndex = 0, args?: any, idOverride: string = undefined, setNameForArrowKeyNav: string = undefined) {
+	public enableInvoke(handleMethod: Function, tabIndex = 0, args?: any, idOverride: string = undefined, setNameForArrowKeyNav: string = undefined, defaultAction: string = undefined) {
 		// Because of the way mithril does the callbacks, we need to rescope it so that "this" points to the class
 		if (handleMethod) {
 			handleMethod = handleMethod.bind(this, args);
@@ -89,25 +89,15 @@ export abstract class ComponentBase<TState, TProps> {
 			},
 			onkeyup: (e: KeyboardEvent) => {
 				let element = e.currentTarget as HTMLElement;
-				if (e.which === Constants.KeyCodes.enter) {
-					// Hitting Enter on <a> tags that contains an href automatically fire the click event, so don't do it again
-					if (!(element.tagName === "A" && element.hasAttribute("href"))) {
-						// Intentionally sending click event before handling the method
-						// TODO replace this comment with a test that validates the call order is correct
-						let id = element.id;
-
-						Clipper.logger.logClickEvent(id);
-
-						if (handleMethod) {
-							handleMethod(e);
-						}
-					}
+				if (e.which === Constants.KeyCodes.space) {
+					ComponentBase.triggerActionEvent(element, handleMethod, e);
 				} else if (e.which === Constants.KeyCodes.tab) {
 					// Since they are using the keyboard, revert to the default value of the outline so it is visible
 					element.style.outlineStyle = "";
 				}
 
-				if (!setNameForArrowKeyNav) {
+				if (!setNameForArrowKeyNav && !defaultAction) {
+					ComponentBase.enterKeySelection(e);
 					return;
 				} else if (element.hasAttribute("data-" + Constants.AriaSet.setNameForArrowKeyNav)) {
 					let posInSet = parseInt(element.getAttribute("aria-posinset"), 10);
@@ -124,6 +114,10 @@ export abstract class ComponentBase<TState, TProps> {
 						}
 						let nextPosInSet = posInSet + 1;
 						ComponentBase.focusOnButton(setNameForArrowKeyNav, nextPosInSet);
+					} else if (!defaultAction) {
+						ComponentBase.enterKeySelection(e);
+					} else if (e.which === Constants.KeyCodes.enter) {
+						ComponentBase.triggerActionEvent(element, handleMethod, e);
 					}
 				}
 			},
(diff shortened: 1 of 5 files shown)


src/scripts/clipperUI/panels/successPanel.tsx
```diff
@@ -38,7 +31,7 @@ class SuccessPanelClass extends ComponentBase<{ }, ClipperStateProp> {
 						{Localization.getLocalizedString("WebClipper.Label.ClipSuccessful")}
 					</span>
 				</div>
-				<a id={Constants.Ids.launchOneNoteButton} {...this.enableInvoke(this.onLaunchOneNoteButton, 70) }>
+				<a id={Constants.Ids.launchOneNoteButton} {...this.enableInvoke(this.onLaunchOneNoteButton, 70, undefined, undefined, undefined, Constants.DefaultAction.clipSuccessPanel) }>
```
mttwc, line 34: Is it intentional for the default action to be a panel? Shouldn't it be a button?

src/scripts/clipperUI/panels/successPanel.tsx
```diff
@@ -1,18 +1,11 @@
 import {Constants} from "../../constants";
```
mttwc, line 1: There are other panels that we need to figure out default actions for. If you look for tsx files with the name 'panel' you will find them

src/scripts/clipperUI/componentBase.ts
```diff
@@ -132,10 +126,36 @@ export abstract class ComponentBase<TState, TProps> {
...
+			let defaultActionElement = document.querySelector("[data-defaultaction]:not([data-defaultaction='undefined'])");
+			let defaultActionElementHtml = defaultActionElement as HTMLElement;
+			if (!!defaultActionElementHtml) {
+				defaultActionElementHtml.click();
+			}
+		}
+	}
+
+	private static triggerActionEvent(element, handleMethod: Function, e: KeyboardEvent) {
+// Hitting Enter on <a> tags that contains an href automatically fire the click event, so don't do it again
```
mttwc, line 145: nit: indentation

src/scripts/clipperUI/componentBase.ts
```diff
@@ -89,25 +89,15 @@ export abstract class ComponentBase<TState, TProps> {
...
+				if (e.which === Constants.KeyCodes.space) {
+					ComponentBase.triggerActionEvent(element, handleMethod, e);
 				} else if (e.which === Constants.KeyCodes.tab) {
 					// Since they are using the keyboard, revert to the default value of the outline so it is visible
 					element.style.outlineStyle = "";
 				}
 
-				if (!setNameForArrowKeyNav) {
+				if (!setNameForArrowKeyNav && !defaultAction) {
+					ComponentBase.enterKeySelection(e);
```
mttwc, line 100: Why do we handle the enter key differently in 3 different places? Once here, and another (albeit differently) in lines 118 and 120?

src/scripts/clipperUI/componentBase.ts
```diff
@@ -124,6 +114,10 @@ export abstract class ComponentBase<TState, TProps> {
 						}
 						let nextPosInSet = posInSet + 1;
 						ComponentBase.focusOnButton(setNameForArrowKeyNav, nextPosInSet);
+					} else if (!defaultAction) {
```
mttwc, line 117: See above comment

src/scripts/clipperUI/componentBase.ts
```diff
@@ -132,10 +126,36 @@ export abstract class ComponentBase<TState, TProps> {
 				element.style.outlineStyle = "none";
 			},
 			tabIndex: tabIndex,
-			"data-setnameforarrowkeynav": setNameForArrowKeyNav
+			"data-setnameforarrowkeynav": setNameForArrowKeyNav,
+			"data-defaultaction": defaultAction
```
mttwc, line 130: If setNameForArrowKeyNav/defaultAction is undefined, do the attributes still get set, but with undefined values? We should make sure these attributes do not get set if they are undefined.

src/scripts/clipperUI/componentBase.ts
```diff
@@ -132,10 +126,36 @@ export abstract class ComponentBase<TState, TProps> {
...
 		};
 	}
 
+	private static enterKeySelection(e: KeyboardEvent) {
+		if (e.which === Constants.KeyCodes.enter) {
+			let defaultActionElement = document.querySelector("[data-defaultaction]:not([data-defaultaction='undefined'])");
```
mttwc, line 136: I was testing other web pages and noticed that sometimes, hitting the enter key on certain elements will behave the same hitting space. For example, in www.google.com, hitting enter on the About page triggers the About page, not the Search. Can you please doublecheck with Ailsa and Valentin on what the appropriate behavior should be?

src/scripts/clipperUI/componentBase.ts
```diff
@@ -67,7 +67,7 @@ export abstract class ComponentBase<TState, TProps> {
 	 * Example use:
 	 *      <a id="myCoolButton" {...this.enableInvoke(this.myButtonHandler, 0)}>Click Me</a>
 	 */
-	public enableInvoke(handleMethod: Function, tabIndex = 0, args?: any, idOverride: string = undefined, setNameForArrowKeyNav: string = undefined) {
+	public enableInvoke(handleMethod: Function, tabIndex = 0, args?: any, idOverride: string = undefined, setNameForArrowKeyNav: string = undefined, defaultAction: string = undefined) {
```
mttwc, line 70: If the user is not a clickable element and hits enter, nothing will happen. We should double check with Ailsa and Valentin on what is the right thing to do in this case.

src/scripts/clipperUI/componentBase.ts
```diff
@@ -67,7 +67,7 @@ export abstract class ComponentBase<TState, TProps> {
 	 * Example use:
```
mttwc, line 67: We need to update the method comment

src/scripts/clipperUI/componentBase.ts
```diff
@@ -89,25 +89,15 @@ export abstract class ComponentBase<TState, TProps> {
...
-					}
+				if (e.which === Constants.KeyCodes.space) {
+					ComponentBase.triggerActionEvent(element, handleMethod, e);
 				} else if (e.which === Constants.KeyCodes.tab) {
 					// Since they are using the keyboard, revert to the default value of the outline so it is visible
 					element.style.outlineStyle = "";
 				}
 
-				if (!setNameForArrowKeyNav) {
+				if (!setNameForArrowKeyNav && !defaultAction) {
```
mttwc, line 99: The existance of setNameForArrowKeyNav and defaultAction seem to be tightly coupled in this method. Do they really need to be?

src/scripts/clipperUI/componentBase.ts
mttwc, line 136: the string "defaultaction" in "data-defaultaction" is duplicated in a bunch of places

mttwc (changes requested):

PR #400, opened 2018-05-29
add aria sets to augmentation options

JaneHazen:


src/scripts/clipperUI/components/previewViewer/previewViewerAugmentationHeader.tsx
```diff
@@ -52,23 +52,25 @@ class PreviewViewerAugmentationHeaderClass extends PreviewViewerHeaderComponentB
...
 				<button
 					aria-label={Localization.getLocalizedString("WebClipper.Accessibility.ScreenReader.ChangeFontToSansSerif")}
-					aria-pressed={!this.props.serif ? "true" : "false"}
+					aria-checked={!this.props.serif ? "true" : "false"}
 					id={Constants.Ids.sansSerif}
-					{...this.enableInvoke(this.props.changeFontFamily, 101, false) }
+					{...this.enableInvoke(this.props.changeFontFamily, !this.props.serif ? 101 : undefined, false, undefined, Constants.AriaSet.serifGroupSet) }
 					className={!this.props.serif ? HeaderClasses.Button.activeControlButton : HeaderClasses.Button.controlButton}
-					type="button">
+					role="radio">
```
mttwc: Rather than change the role to radio, we should change this element from a button to an input element of type="radio".

src/scripts/clipperUI/components/previewViewer/previewViewerAugmentationHeader.tsx
```diff
@@ -52,23 +52,25 @@ class PreviewViewerAugmentationHeaderClass extends PreviewViewerHeaderComponentB
 	private getSerifGroup(): ControlGroup {
 		return {
 			id: Constants.Ids.serifControl,
+			role: "radiogroup",
+			isAriaSet: true,
 			innerElements: [
 				<button
 					aria-label={Localization.getLocalizedString("WebClipper.Accessibility.ScreenReader.ChangeFontToSansSerif")}
-					aria-pressed={!this.props.serif ? "true" : "false"}
+					aria-checked={!this.props.serif ? "true" : "false"}
```
mttwc: We can shorten this by using !this.props.serif + ""

src/scripts/clipperUI/components/previewViewer/previewViewerAugmentationHeader.tsx
```diff
@@ -52,23 +52,25 @@ class PreviewViewerAugmentationHeaderClass extends PreviewViewerHeaderComponentB
...
-					{...this.enableInvoke(this.props.changeFontFamily, 101, false) }
+					{...this.enableInvoke(this.props.changeFontFamily, !this.props.serif ? 101 : undefined, false, undefined, Constants.AriaSet.serifGroupSet) }
 					className={!this.props.serif ? HeaderClasses.Button.activeControlButton : HeaderClasses.Button.controlButton}
-					type="button">
+					role="radio">
 					{Localization.getLocalizedString("WebClipper.Preview.Header.SansSerifButtonLabel") }
 				</button>,
 				<button
 					aria-label={Localization.getLocalizedString("WebClipper.Accessibility.ScreenReader.ChangeFontToSerif")}
-					aria-pressed={this.props.serif ? "true" : "false"}
```
mttwc, line 67: See above comment on the shorthand

src/scripts/clipperUI/components/previewViewer/previewViewerAugmentationHeader.tsx
```diff
@@ -52,23 +52,25 @@ class PreviewViewerAugmentationHeaderClass extends PreviewViewerHeaderComponentB
...
+					aria-checked={this.props.serif ? "true" : "false"}
 					id={Constants.Ids.serif}
-					{...this.enableInvoke(this.props.changeFontFamily, 102, true) }
+					{...this.enableInvoke(this.props.changeFontFamily, this.props.serif ? 101 : undefined, true, undefined, Constants.AriaSet.serifGroupSet) }
```
mttwc: This `101` value should be a constant because it's duplicated in two places

src/scripts/clipperUI/components/previewViewer/previewViewerAugmentationHeader.tsx
```diff
@@ -79,16 +81,17 @@ class PreviewViewerAugmentationHeaderClass extends PreviewViewerHeaderComponentB
...
 					aria-label={Localization.getLocalizedString("WebClipper.Accessibility.ScreenReader.DecreaseFontSize")}
-					type="button" {...this.enableInvoke(this.props.changeFontSize, 103, false) }
+					type="button" {...this.enableInvoke(this.props.changeFontSize, 103, false, undefined, Constants.AriaSet.fontSizeSet) }
 					id={Constants.Ids.decrementFontSize}>
 					<img src={ExtensionUtils.getImageResourceUrl("editorOptions/font_down.png") } />
 				</button>,
 				<button className={HeaderClasses.Button.controlButton}
 					aria-label={Localization.getLocalizedString("WebClipper.Accessibility.ScreenReader.IncreaseFontSize")}
-					type="button" {...this.enableInvoke(this.props.changeFontSize, 104, true) }
+					type="button" {...this.enableInvoke(this.props.changeFontSize, undefined, true, undefined, Constants.AriaSet.fontSizeSet) }
```
mttwc: Since these are buttons, is it intended that you removed the tabIndex here?

src/scripts/clipperUI/components/previewViewer/previewViewerHeaderComponentBase.tsx
```diff
@@ -40,8 +40,18 @@ export abstract class PreviewViewerHeaderComponentBase<T, P> extends ComponentBa
 		for (let i = 0; i < buttonGroups.length; i++) {
 			let id = buttonGroups[i].id;
 			let className = buttonGroups[i].className;
+			let role = buttonGroups[i].role;
+			let isAriaSet = buttonGroups[i].isAriaSet;
+			if (isAriaSet) {
+				let setsize = buttonGroups[i].innerElements.length;
```
mttwc: Name should follow convention `setSize`

src/scripts/clipperUI/components/previewViewer/previewViewerHeaderComponentBase.tsx
```diff
@@ -40,8 +40,18 @@ export abstract class PreviewViewerHeaderComponentBase<T, P> extends ComponentBa
 		for (let i = 0; i < buttonGroups.length; i++) {
 			let id = buttonGroups[i].id;
```
mttwc: Recommend assigning buttonGroups[i] to a variable so it's not repeated everywhere

src/scripts/constants.ts
```diff
@@ -436,5 +436,7 @@ export module Constants {
 	export module AriaSet {
 		export var setNameForArrowKeyNav = "setnameforarrowkeynav";
```
mttwc: I think the AriaSet module is doing more than one thing here, i.e., setNameForArrowKeyNav refers to an attribute name, while the other values are set names. I recommend keeping these in different modules.

src/tests/assert.ts
```diff
@@ -20,8 +20,10 @@ export module Assert {
 
 		// Assert tab index ordering
 		for (let i = 1; i < elements.length; i++) {
-			ok(elements[i].elem.tabIndex > elements[i - 1].elem.tabIndex,
-				"Element " + elements[i].name + " should have a greater tabIndex than element " + elements[i - 1].name);
+			if (elements[i].elem.tabIndex) {
```
mttwc: Why was this added? Wouldn't this give a false positive on tests where we expect an element to have tabindexes but doesn't?

src/tests/clipperUI/components/previewViewer/augmentationPreview_tests.tsx
```diff
@@ -38,6 +38,15 @@ export class AugmentationPreviewTests extends TestModule {
 				Constants.Ids.incrementFontSize, Constants.Ids.previewHeaderInput]);
 		});
 
+		test("The aria-posinset attribute should flow in order, assuming they are all available", () => {
+			let mockClipperState = this.getMockAugmentationModeState();
+			let defaultComponent = <AugmentationPreview clipperState={mockClipperState} />;
+			MithrilUtils.mountToFixture(defaultComponent);
+
+			Assert.posInSetIsDecremental([Constants.Ids.sansSerif, Constants.Ids.serif]);
```
mttwc: Can we merge these two so we can test tabbing between these two sets?

mttwc (changes requested):

src/scripts/clipperUI/components/previewViewer/previewViewerAugmentationHeader.tsx
```diff
@@ -52,23 +52,25 @@ class PreviewViewerAugmentationHeaderClass extends PreviewViewerHeaderComponentB
...
+			role: "radiogroup",
+			isAriaSet: true,
 			innerElements: [
 				<button
 					aria-label={Localization.getLocalizedString("WebClipper.Accessibility.ScreenReader.ChangeFontToSansSerif")}
-					aria-pressed={!this.props.serif ? "true" : "false"}
+					aria-checked={!this.props.serif + ""}
 					id={Constants.Ids.sansSerif}
-					{...this.enableInvoke(this.props.changeFontFamily, 101, false) }
+					{...this.enableInvoke(this.props.changeFontFamily, !this.props.serif ? Constants.ConditionalTabIndices.oneHundredAndOne : undefined, false, undefined, Constants.AriaSet.serifGroupSet) }
```
mttwc: Sorry, I meant a constant in this class or method. It's overkill to have this declared in Constants.ts as every file will then have access to it.

src/scripts/clipperUI/components/previewViewer/previewViewerAugmentationHeader.tsx
```diff
@@ -52,23 +52,25 @@ class PreviewViewerAugmentationHeaderClass extends PreviewViewerHeaderComponentB
...
+					role="radio">
 					{Localization.getLocalizedString("WebClipper.Preview.Header.SansSerifButtonLabel") }
 				</button>,
 				<button
 					aria-label={Localization.getLocalizedString("WebClipper.Accessibility.ScreenReader.ChangeFontToSerif")}
-					aria-pressed={this.props.serif ? "true" : "false"}
+					aria-checked={this.props.serif + ""}
 					id={Constants.Ids.serif}
-					{...this.enableInvoke(this.props.changeFontFamily, 102, true) }
+					{...this.enableInvoke(this.props.changeFontFamily, this.props.serif ? Constants.ConditionalTabIndices.oneHundredAndOne : undefined, true, undefined, Constants.AriaSet.serifGroupSet) }
```
mttwc: See above

src/scripts/clipperUI/components/previewViewer/previewViewerHeaderComponentBase.tsx
```diff
@@ -38,11 +38,22 @@ export abstract class PreviewViewerHeaderComponentBase<T, P> extends ComponentBa
 		let buttonGroups = this.getControlGroups();
 
 		for (let i = 0; i < buttonGroups.length; i++) {
-			let id = buttonGroups[i].id;
-			let className = buttonGroups[i].className;
+			let currentButtonGroup = buttonGroups[i];
```
mttwc: You may be able to use this shorthand syntax:

const { id, className, role, isAriaSet } = currentButtonGroup

src/tests/assert.ts
```diff
@@ -20,9 +20,9 @@ export module Assert {
 
 		// Assert tab index ordering
 		for (let i = 1; i < elements.length; i++) {
-			ok(elements[i].elem.tabIndex > elements[i - 1].elem.tabIndex,
-				"Element " + elements[i].name + " should have a greater tabIndex than element " + elements[i - 1].name);
-		}
+				ok(elements[i].elem.tabIndex > elements[i - 1].elem.tabIndex,
```
mttwc: Fix indentation

src/tests/clipperUI/components/previewViewer/augmentationPreview_tests.tsx
```diff
@@ -34,10 +30,27 @@ export class AugmentationPreviewTests extends TestModule {
...
+			Assert.tabOrderIsIncremental([Constants.Ids.highlightButton, Constants.Ids.sansSerif, Constants.Ids.decrementFontSize,
 				Constants.Ids.incrementFontSize, Constants.Ids.previewHeaderInput]);
 		});
 
+		test("The aria-posinset attribute should flow in order, assuming they are all available", () => {
+			let mockClipperState = this.getMockAugmentationModeState();
+			let defaultComponent = <AugmentationPreview clipperState={mockClipperState} />;
+			MithrilUtils.mountToFixture(defaultComponent);
+
+			Assert.posInSetIsDecremental([Constants.Ids.sansSerif, Constants.Ids.serif]);
```
mttwc: I know posInSetIsDecremental was made in a different PR, but is there any reason why we are testing decrementally than incrementally?

src/tests/clipperUI/components/previewViewer/augmentationPreview_tests.tsx
```diff
@@ -34,10 +30,27 @@ export class AugmentationPreviewTests extends TestModule {
 			let defaultComponent = <AugmentationPreview clipperState={mockClipperState} />;
 			MithrilUtils.mountToFixture(defaultComponent);
 
-			Assert.tabOrderIsIncremental([Constants.Ids.highlightButton, Constants.Ids.sansSerif, Constants.Ids.serif, Constants.Ids.decrementFontSize,
+			MithrilUtils.simulateAction(() => {
+				document.getElementById(Constants.Ids.serif).click();
+			});
+			Assert.tabOrderIsIncremental([Constants.Ids.highlightButton, Constants.Ids.serif, Constants.Ids.decrementFontSize,
```
mttwc, line 36: As I mentioned, we would want to add tests to ensure that the tabindex does not get set on elements that shouldn't have them. I think these tests will pass if the tabindex gets set on both sans-serif and serif buttons, when they should fail.

src/scripts/constants.ts
```diff
@@ -433,8 +433,17 @@ export module Constants {
...
+	}
+
+	export module AriaSet {
 		export var modeButtonSet = "ariaModeButtonSet";
+		export var serifGroupSet = "serifGroupSet";
+		export var fontSizeSet = "fontSizeSet";
+	}
+
+	export module ConditionalTabIndices {
```
mttwc: See above comment. Having this const being declared here is overkill.

mttwc (changes requested):

src/tests/assert.ts
```diff
@@ -20,17 +20,16 @@ export module Assert {
 
 		// Assert tab index ordering
 		for (let i = 1; i < elements.length; i++) {
-			ok(elements[i].elem.tabIndex > elements[i - 1].elem.tabIndex,
-				"Element " + elements[i].name + " should have a greater tabIndex than element " + elements[i - 1].name);
-		}
+				ok(elements[i].elem.tabIndex > elements[i - 1].elem.tabIndex, "Element " + elements[i].name + " whose tabIndex is" + elements[i].elem.tabIndex + " should have a greater tabIndex than element " + elements[i - 1].name + " whose tabIndex is " + elements[i - 1].elem.tabIndex);
```
mttwc: Fix indentation

src/tests/assert.ts
```diff
@@ -20,17 +20,16 @@ export module Assert {
...
+			}
 	}
 
-	export function posInSetIsDecremental(elementIds: string[]) {
+	export function posInSetIsIncremental(elementIds: string[]) {
```
mttwc: Was it testing incrementally already, but it was just the name that was incorrect?

src/tests/clipperUI/components/modeButtonSelector_tests.tsx
```diff
@@ -203,8 +203,6 @@ export class ModeButtonSelectorTests extends TestModule {
 		});
 
 		test("Other modes' buttons should have the 'selected' class styling applied to it if they are clicked on", () => {
-			let controllerInstance = MithrilUtils.mountToFixture(this.defaultComponent);
```
mttwc: Why were these lines removed?

src/tests/clipperUI/components/previewViewer/augmentationPreview_tests.tsx
```diff
@@ -34,10 +30,49 @@ export class AugmentationPreviewTests extends TestModule {
...
 
+		test("When a pair of codependent buttons have conditional attributes, they both respond with the same behavior when clicked and have different values from each other", () => {
+			let mockClipperState = this.getMockAugmentationModeState();
+			let defaultComponent = <AugmentationPreview clipperState={mockClipperState} />;
+			MithrilUtils.mountToFixture(defaultComponent);
+
+			MithrilUtils.simulateAction(() => {
+				document.getElementById(Constants.Ids.serif).click();
+			});
+			let selectedTags = [document.getElementById(Constants.Ids.serif).tabIndex];
```
mttwc: Variable names could be a little earlier on what they represent (they currently do not make any reference to the tabIndex)

src/tests/clipperUI/components/previewViewer/augmentationPreview_tests.tsx
```diff
@@ -34,10 +30,49 @@ export class AugmentationPreviewTests extends TestModule {
...
+			let unselectedTags = [document.getElementById(Constants.Ids.sansSerif).tabIndex];
+
+			MithrilUtils.simulateAction(() => {
+				document.getElementById(Constants.Ids.sansSerif).click();
+			});
+			selectedTags.push(document.getElementById(Constants.Ids.sansSerif).tabIndex);
+			unselectedTags.push(document.getElementById(Constants.Ids.serif).tabIndex);
+
+			strictEqual(selectedTags[0], selectedTags[1], "When the serif button is clicked the tabIndex should be the same as the tabIndex of the sansSerif button when clicked");
+			strictEqual(unselectedTags[0], unselectedTags[1], "When the serif button is not clicked the tabIndex should be the same as the tabIndex of the sansSerif button when not clicked");
```
mttwc: I'd explicitly check that these have values of -1 to ensure they are not tabbable

mttwc: Similarly check that the expected tabIndexes have positive values

src/tests/clipperUI/components/previewViewer/augmentationPreview_tests.tsx
```diff
@@ -34,10 +30,49 @@ export class AugmentationPreviewTests extends TestModule {
...
+			notStrictEqual(selectedTags[0], unselectedTags[0], "The buttons should have different values when one is clicked");
+		});
+
+		test("The aria-posinset attribute should flow in order, assuming they are all available", () => {
+			let mockClipperState = this.getMockAugmentationModeState();
+			let defaultComponent = <AugmentationPreview clipperState={mockClipperState} />;
+			MithrilUtils.mountToFixture(defaultComponent);
+
+			Assert.posInSetIsIncremental([Constants.Ids.sansSerif, Constants.Ids.serif]);
```
mttwc: This is not part of your current change, but I don't think we should test posinset in this way. For example, if I had 2 set elements with posinset values of 1 and 3 respectively, it's incorrect as I expect them to be 1 and 2. We should change these posinset tests to explicitly check for specific numbers.

mttwc (changes requested):

src/tests/assert.ts
JaneHazen: Yeah.... Sorry!

src/tests/clipperUI/components/modeButtonSelector_tests.tsx
JaneHazen: The variable wasn't being used anywhere

src/tests/clipperUI/components/previewViewer/augmentationPreview_tests.tsx
```diff
@@ -29,13 +25,64 @@ export class AugmentationPreviewTests extends TestModule {
 	}
 
 	protected tests() {
+		function isPositive(value: number): number {
```
mttwc: This is not needed:
a) You can just assert n > 0 in the test itself
b) Why does this return a number and not a boolean?
c) 0 is a valid positive tabindex but is treated as negative here

src/tests/assert.ts
```diff
@@ -20,22 +20,22 @@ export module Assert {
 
 		// Assert tab index ordering
 		for (let i = 1; i < elements.length; i++) {
-			ok(elements[i].elem.tabIndex > elements[i - 1].elem.tabIndex,
-				"Element " + elements[i].name + " should have a greater tabIndex than element " + elements[i - 1].name);
-		}
```
mttwc: Fix indentation

src/tests/clipperUI/components/previewViewer/augmentationPreview_tests.tsx
```diff
@@ -29,13 +25,64 @@ export class AugmentationPreviewTests extends TestModule {
...
+			});
+			Assert.tabOrderIsIncremental([Constants.Ids.highlightButton, Constants.Ids.sansSerif, Constants.Ids.decrementFontSize,
+				Constants.Ids.incrementFontSize, Constants.Ids.previewHeaderInput]);
+		});
+
+		test("When a pair of codependent buttons have conditional attributes, they both respond with the same behavior when clicked and have different values from each other", () => {
+			let mockClipperState = this.getMockAugmentationModeState();
+			let defaultComponent = <AugmentationPreview clipperState={mockClipperState} />;
+			MithrilUtils.mountToFixture(defaultComponent);
+			let unSelectedTabShouldBeNegative = -1;
```
mttwc: No need to declare a variable for this. It's easier to read strictEqual(unselectedTabIndices[0], -1, ...

src/tests/clipperUI/components/previewViewer/augmentationPreview_tests.tsx
```diff
@@ -29,13 +25,64 @@ export class AugmentationPreviewTests extends TestModule {
...
+			selectedTabIndices.push(document.getElementById(Constants.Ids.sansSerif).tabIndex);
+			unselectedTabIndices.push(document.getElementById(Constants.Ids.serif).tabIndex);
+
+			strictEqual(selectedTabIndices[0], selectedTabIndices[1], "When the serif button is clicked the tabIndex should be the same as the tabIndex of the sansSerif button when clicked");
+			ok(isPositive(selectedTabIndices[0]), "Selected tabIndex should be greater than 0");
+
+			strictEqual(unselectedTabIndices[0], unselectedTabIndices[1], "When the serif button is not clicked the tabIndex should be the same as the tabIndex of the sansSerif button when not clicked");
+			strictEqual(unselectedTabIndices[0], unSelectedTabShouldBeNegative, "The unselected tabsIndices should be -1");
+
+			notStrictEqual(selectedTabIndices[0], unselectedTabIndices[0], "The buttons should have different values when one is clicked");
```
mttwc: This line is not necessary since we already asserted the first value is positive and the second value is negative

src/tests/assert.ts
```diff
@@ -20,22 +20,22 @@ export module Assert {
...
-		Assert.posInSetIsDecrementalForElements(elementsInExpectedPosInSetOrder);
+		Assert.posInSetIsIncrementalForElements(elementsInExpectedPosInSetOrder);
 	}
 
-	export function posInSetIsDecrementalForElements(elements: {name: string, elem: HTMLElement}[]) {
+	export function posInSetIsIncrementalForElements(elements: {name: string, elem: HTMLElement}[]) {
 		if (elements.length < 2 ) {
 			return;
 		}
 		for (let i = 0; i < (elements.length - 1 ); i++) {
```
mttwc: You should be testing the last element of the list too

src/tests/assert.ts
```diff
@@ -20,22 +20,22 @@ export module Assert {
...
-			ok(elements[i].elem.getAttribute("aria-posinset") < elements[i + 1].elem.getAttribute("aria-posinset"), "Element " + elements[i].name + "should have a posInSet less than " + elements[i + 1].name );
+			let expected = (i + 1).toString();
+			strictEqual(elements[i].elem.getAttribute("aria-posinset"), expected, "Element" + elements[i].name + "has the wrong aria-posinset value");
```
mttwc: The test message should have spaces around elements[i].name

src/tests/assert.ts
```diff
@@ -20,22 +20,22 @@ export module Assert {
...
 
-	export function posInSetIsDecremental(elementIds: string[]) {
+	export function posInSetIsIncremental(elementIds: string[]) {
 		let elementsInExpectedPosInSetOrder = elementIds.map(id => {return{name: id, elem: document.getElementById(id) }; });
-		Assert.posInSetIsDecrementalForElements(elementsInExpectedPosInSetOrder);
+		Assert.posInSetIsIncrementalForElements(elementsInExpectedPosInSetOrder);
 	}
 
-	export function posInSetIsDecrementalForElements(elements: {name: string, elem: HTMLElement}[]) {
+	export function posInSetIsIncrementalForElements(elements: {name: string, elem: HTMLElement}[]) {
```
mttwc: Test name could be more descriptive. We're not just testing that posinset is incremental, but it follows a strict order of 1, 2 ... n

src/tests/assert.ts
mttwc: We should also test for setsize.

mttwc (changes requested):

src/tests/assert.ts
```diff
@@ -20,22 +21,22 @@ export module Assert {
...
 		// Assert tab index ordering
 		for (let i = 1; i < elements.length; i++) {
-			ok(elements[i].elem.tabIndex > elements[i - 1].elem.tabIndex,
-				"Element " + elements[i].name + " should have a greater tabIndex than element " + elements[i - 1].name);
+			ok(elements[i].elem.tabIndex > elements[i - 1].elem.tabIndex, "Element " + elements[i].name + " whose tabIndex is" + elements[i].elem.tabIndex + " should have a greater tabIndex than element " + elements[i - 1].name + " whose tabIndex is " + elements[i - 1].elem.tabIndex);
 		}
 	}
 
-	export function posInSetIsDecremental(elementIds: string[]) {
+	export function checkAriaGroupingAttributes(elementIds: string[]) {
```
mttwc: Use term 'set' instead of 'group'

src/tests/clipperUI/components/modeButtonSelector_tests.tsx
```diff
@@ -163,11 +163,23 @@ export class ModeButtonSelectorTests extends TestModule {
 
 		test("The aria-posinset attribute should flow in element order, assuming they are all available", () => {
 			let startingState = MockProps.getMockClipperState();
+
+			startingState.invokeOptions.invokeMode = InvokeMode.ContextImage;
```
mttwc: Separate these into different tests to minimize risk of tests not being pure

mttwc (changes requested):

mttwc (approved):

PR #405, opened 2018-06-05
add home and end key navigation

JaneHazen:


src/scripts/clipperUI/componentBase.ts
```diff
@@ -136,12 +142,12 @@ export abstract class ComponentBase<TState, TProps> {
 		};
 	}
 
-	private static focusOnButton(setNameForArrowKeyNav: string, nextPosInSet: number) {
+	private static focusOnButton(setNameForArrowKeyNav: string, stoppingPoint: number) {
```
mttwc: I prefer the old name (nextPosInSet) as its more descriptive. Do someone who doesn't have access to the code of this method, it's not clear what `stoppingPoint` does

mttwc (approved):

PR #413, opened 2018-06-12
fix scrolling with up and down arrows

JaneHazen:


src/scripts/clipperUI/componentBase.ts
```diff
@@ -87,6 +87,7 @@ export abstract class ComponentBase<TState, TProps> {
 
 		invokeAttributes.onkeyup = (e: KeyboardEvent) => {
 			e.preventDefault();
+			e.stopImmediatePropagation();
```
mttwc, line 90: Small fixes are the best :)

mttwc (approved):

PR #416, opened 2018-06-14
css

JaneHazen:


src/styles/clipper.less
```diff
@@ -155,7 +155,7 @@
 
 #regionClipCancelButton:focus {
 	outline-offset: 4px;
-	outline: ActiveBorder solid 1px !important;
+	outline: @FocusOnPurpleBackground solid 1px !important;
```
mttwc, line 158: I like this variable name (it explains the 'why' very well!)

mttwc (approved):

PR #419, opened 2018-06-19
css fixes

JaneHazen:


src/scripts/clipperUI/componentBase.ts
```diff
@@ -158,6 +158,12 @@ export abstract class ComponentBase<TState, TProps> {
 					element.style.outlineStyle = "";
 				}
 			},
+			onkeydown: (e: KeyboardEvent) => {
+				if (e.which === Constants.KeyCodes.space) {
```
mttwc: If you press space in a text box, would this prevent the space character from being inputted?

src/styles/clipper.less
```diff
@@ -76,14 +76,13 @@
 }
 
 #closeButtonContainer {
-	position: absolute;
+	float: right;
```
mttwc, line 79: What does this do?

src/styles/mainClassDefinitions.less
```diff
@@ -467,8 +461,8 @@
 	border-radius: 4px;
 	text-align: center;
 	margin-bottom: 10px;
-	margin-left: 5px;
 	margin-right: 5px;
+	width: 100%;
```
mttwc: Is this for the clip button? Wouldn't leave in the margin-left cause this to not take up all the space on the right?

mttwc (changes requested):

src/styles/clipper.less
JaneHazen: Using absolute positioning makes the CSS a mess for all of its nested elements, so I just floated this to the right to get it in the right position without making the closeButton and closeButtonIcon do crazy thing

src/scripts/clipperUI/componentBase.ts
JaneHazen: That still works fine- when it is in the textbox it still allows spaces and enter for paragraphs, this just directs you into that box

src/styles/mainClassDefinitions.less
JaneHazen: I'm not sure what you mean? The margin-left was causing this to pop out of alignment with the location label and the picker

mttwc: As discussed, we should be able to remove margin-right

src/scripts/clipperUI/componentBase.ts
mttwc: Confirmed working

src/styles/mainClassDefinitions.less
```diff
@@ -467,8 +461,8 @@
 	border-radius: 4px;
 	text-align: center;
 	margin-bottom: 10px;
-	margin-left: 5px;
```
mttwc, line 470: As discussed, when this comes into focus, the UI shifts a little.

mttwc (changes requested):

src/styles/mainClassDefinitions.less
```diff
@@ -462,7 +462,7 @@
 	text-align: center;
 	margin-bottom: 10px;
 	margin-right: 5px;
-	width: 100%;
+	margin-left: 5px;
```
mttwc: As discussed, we need to consult with design to see if we'd rather have missing horizontal focus borders or misaligned elements. Either way, we need to deal with a tradeoff and we try not to make these decisions independently as they will no longer match redlines.

mttwc (changes requested):

mttwc (approved):

PR #421, opened 2018-06-20
take out horizontal arrow navigation

JaneHazen:


src/scripts/clipperUI/components/previewViewer/previewViewerAugmentationHeader.tsx
```diff
@@ -68,7 +68,7 @@ class PreviewViewerAugmentationHeaderClass extends PreviewViewerHeaderComponentB
 					aria-label={Localization.getLocalizedString("WebClipper.Accessibility.ScreenReader.ChangeFontToSerif")}
 					aria-checked={this.props.serif + ""}
 					id={Constants.Ids.serif}
-					{...this.enableAriaInvoke({callback: this.props.changeFontFamily, tabIndex: this.props.serif ? tabIndexOn : tabIndexOff, args: true, ariaSetName: Constants.AriaSet.serifGroupSet, ariaSetDirection: AriaNavDirection.Horizontal})}
```
mttwc, line 71: Sorry for the confusion. The requirement is that left/right behaves the same as up/down, not that we remove horizontal capabilities. I.e., in a listbox widget, left does the same thing as up, and right does the same thing as down

mttwc (changes requested):

PR #434, opened 2019-05-17
Make focus indicator clearly visible for high contrast pages

jogonzal:
image
Fix for https://office.visualstudio.com/OneNote/_workitems/edit/3174504

### DIFF
--- a/src/styles/globalStyles.less
+++ b/src/styles/globalStyles.less
@@ -101,5 +101,5 @@ a {
 }
 
 :focus {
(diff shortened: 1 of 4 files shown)


src/styles/globalStyles.less
```diff
@@ -101,5 +101,5 @@ a {
 }
 
 :focus {
-	outline: @FocusOnPurpleBackground auto 1px;
```
jogonzal, line 104: auto is not a valid value, so changing to solid

mttwc (approved):

PR #435, opened 2019-05-17
View in OneNote button should be focusable

jogonzal:
image
Fix for https://office.visualstudio.com/OneNote/_workitems/edit/3174756

### DIFF
--- a/src/scripts/clipperUI/panels/dialogPanel.tsx
+++ b/src/scripts/clipperUI/panels/dialogPanel.tsx
@@ -57,10 +57,10 @@ export abstract class DialogPanelClass extends ComponentBase<{}, DialogPanelProp
 							return (
 								<a id={button.id} className="dialogButton" {...this.enableInvoke({callback: button.handler, tabIndex: 70})}>
 									<div className="wideButtonContainer">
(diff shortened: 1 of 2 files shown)


src/scripts/clipperUI/panels/dialogPanel.tsx
```diff
@@ -57,10 +57,10 @@ export abstract class DialogPanelClass extends ComponentBase<{}, DialogPanelProp
 							return (
 								<a id={button.id} className="dialogButton" {...this.enableInvoke({callback: button.handler, tabIndex: 70})}>
 									<div className="wideButtonContainer">
-										<span className="wideButtonFont wideActionButton"
+										<a className="wideButtonFont wideActionButton"
```
jogonzal, line 60: no idea why this was a span instead of an anchor or <button

mttwc: When the Clipper was written, our team made the mistake of not using elements with the appropriate semantic meaning. We generally looked at the default style of the elements, and picked one that would require the least css to get it to what it should look like. That meant we have a lot of divs and spans that should have been other elements like buttons :S

mttwc (approved):

PR #441, opened 2019-05-21
Prevent focus on Edge+Narrator from going into the background page by looping focus on our page

jogonzal:
Fix for: https://office.visualstudio.com/OneNote/_workitems/edit/3180040

### DIFF
--- a/src/scripts/clipperUI/panels/optionsPanel.tsx
+++ b/src/scripts/clipperUI/panels/optionsPanel.tsx
@@ -61,6 +61,25 @@ class OptionsPanelClass extends ComponentBase<{}, OptionsPanelProp> {
 				if (ev.altKey && ev.which === Constants.KeyCodes.c) {
 					this.checkOptionsBeforeStartClip();
 				}
+
+				// Prevent focus on Edge+Narrator from going into the background page by looping focus on our page
+				if (ev.shiftKey && ev.keyCode === Constants.KeyCodes.tab) {
+					const targetAsAny = ev.target as any
+					if (targetAsAny && targetAsAny.id) {
+						const idsThatShouldLoopFocus = [
+							"fullPageButton",
+							"bookmarkButton",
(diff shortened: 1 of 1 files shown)


src/scripts/clipperUI/panels/optionsPanel.tsx
```diff
@@ -61,6 +61,25 @@ class OptionsPanelClass extends ComponentBase<{}, OptionsPanelProp> {
...
 				}
+
+				// Prevent focus on Edge+Narrator from going into the background page by looping focus on our page
+				if (ev.shiftKey && ev.keyCode === Constants.KeyCodes.tab) {
+					const targetAsAny = ev.target as any
+					if (targetAsAny && targetAsAny.id) {
+						const idsThatShouldLoopFocus = [
+							"fullPageButton",
+							"bookmarkButton",
+							"regionButton",
```
mttwc, line 72: Add selectionButton

src/scripts/clipperUI/panels/optionsPanel.tsx
```diff
@@ -61,6 +61,25 @@ class OptionsPanelClass extends ComponentBase<{}, OptionsPanelProp> {
...
+							"augmentationButton"
+						]
+						if (idsThatShouldLoopFocus.indexOf(targetAsAny.id) != -1) {
```
mttwc, line 75: Wouldn't this prevent the user from being able to tab from picker -> clip -> feedback -> signed in?

mttwc (approved):

src/scripts/clipperUI/panels/optionsPanel.tsx
jogonzal: This only affects the "clipping" view where you have the 4 options of clipping, not the signed in view. The sign in view still has a similar bug, but that's on another view.

PR #460, opened 2020-05-20
Fixing -  Button text in high contrast has coloring of normal text

sugupta427:
**Issue** - Button text in high contrast has coloring of normal text. The [bug 3176988](https://office.visualstudio.com/DefaultCollection/OneNote/_workitems/edit/3176988)  here says the text should take link color but the text inside the button is not a link so ideally it should take color of button-text. 

**Fix** - The buttons here are made by <a> with role = "button" instead of <button> tag, this is the reason why high contrast mode is treating the text inside it as normal text.  added css to pickup right color for button text in high contrast. The bug is only for particular button, fixing the same for all the buttons.

**Test** - Tested with all available High contrast modes in windows that the color of button-text is taken up correctly.

### DIFF
--- a/src/scripts/clipperUI/components/footer.tsx
+++ b/src/scripts/clipperUI/components/footer.tsx
@@ -106,16 +106,8 @@ class FooterClass extends ComponentBase<FooterState, FooterProps> {
 							style={Localization.getFontFamilyAsStyle(Localization.FontFamily.Regular)}>
 							<hr className="userDivider" />
 							<div className="userDetails">
-<<<<<<< HEAD
 								<label for={Constants.Ids.signOutButton} id={Constants.Ids.currentUserEmail}>{this.props.clipperState.userResult.data.user.emailAddress}</label>
-								<a id={Constants.Ids.signOutButton} role="button" className="userActionButton"
-=======
-								<label for={Constants.Ids.signOutButton} id={Constants.Ids.currentUserEmail} aria-hidden="true">
-									{this.props.clipperState.userResult.data.user.emailAddress}
-								</label>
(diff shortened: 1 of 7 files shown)


vadobre (approved):

src/scripts/clipperUI/components/footer.tsx
```diff
@@ -86,12 +86,12 @@ class FooterClass extends ComponentBase<FooterState, FooterProps> {
...
 										this.props.clipperState.userResult.data.user.fullName
-										? <div id={Constants.Ids.currentUserName}>{this.props.clipperState.userResult.data.user.fullName}</div>
-										: <div id={Constants.Ids.currentUserName}>{Localization.getLocalizedString("WebClipper.Label.SignedIn")}</div>
+										? <div id={Constants.Ids.currentUserName} class="buttonTextInHighContrast">{this.props.clipperState.userResult.data.user.fullName}</div>
+										: <div id={Constants.Ids.currentUserName} class="buttonTextInHighContrast">{Localization.getLocalizedString("WebClipper.Label.SignedIn")}</div>
 									}
 									{
 										this.props.clipperState.userResult.data.user.emailAddress
-										? <div id={Constants.Ids.currentUserId} className="currentUserIdFont" style={Localization.getFontFamilyAsStyle(Localization.FontFamily.Regular)}>{this.props.clipperState.userResult.data.user.emailAddress}</div>
+										? <div id={Constants.Ids.currentUserId} className="buttonTextInHighContrast currentUserIdFont" style={Localization.getFontFamilyAsStyle(Localization.FontFamily.Regular)}>{this.props.clipperState.userResult.data.user.emailAddress}</div>
```
ksboora, line 94: multiple classes?

PR #467, opened 2020-09-22
Bump to v3.8.2 with old Clipper logo, Remove webRequestBlocking permission

DhruvM7:
Changes:
* Bumped version to 3.8.2
* Reverted to old Clipper logo because we don't have updated images for extension store listings right now
* Removed unused extension permission "webRequestBlocking"

Testing:
Went through ship checklist with output of `npm run build:prod` on Chrome. Shipping on Chrome is 1st priority, we will test on other browsers as we publish to their stores.


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Change Log
 
+## 3.8.2 (September 22, 2020)
+* Improvement: Added a11y improvements
+* Removed unused webRequestBlocking extension permission
```
sugupta427, line 5: Info about switching to old clipper logo should also be given here?

sugupta427: I agree to your point, since the new logo never went to prod, there is no point of adding this info here.

sugupta427 (approved):
- [ ] 

- [ ]

PR #479, opened 2021-06-09
Resolve a11y luminosity issues

manu8996:
* Added a line under previewHeaderContainer to resolve the contrast failure between highlighted text and header container
* Placeholder text for textareas is light grey for all browsers. Changed the tag to input to add a css for placeholder text colour for different browsers.

Verified expected behaviour locally.

[ Bug 4281633](https://office.visualstudio.com/OneNote/_workitems/edit/4281633): [Visual Requirements - Clipper_Highlighted text]: Luminosity ratio is 1:0:1 for the yellow colour of highlighted boundary with white colour.

[ Bug 4274058](https://office.visualstudio.com/OneNote/_workitems/edit/4274058): [Visual Requirements - Clipper_Add a note]: Luminosity ratio is 3:4:1 for "Add a note" placeholder text

### DIFF
--- a/src/scripts/clipperUI/components/annotationInput.tsx
+++ b/src/scripts/clipperUI/components/annotationInput.tsx
@@ -47,7 +47,7 @@ class AnnotationInputClass extends ComponentBase<{}, ClipperStateProp> {
 					</span>
 					<br/>
 				</pre>
-				<input
+				<textarea
 					id={Constants.Ids.annotationField}
 					className={Constants.Classes.textAreaInput}
 					role="textbox"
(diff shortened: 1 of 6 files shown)


src/scripts/clipperUI/components/annotationInput.tsx
```diff
@@ -47,7 +47,7 @@ class AnnotationInputClass extends ComponentBase<{}, ClipperStateProp> {
 					</span>
 					<br/>
 				</pre>
-				<textarea
+				<input
```
varaizad: Is there no way we can change the placeholder color for textarea from css? because changing tags might have other issues as well such as input is a focusable element etc.

manu8996: There's a dead code in css for annotation placeholder and initially I was trying to add the new items there, so didn't check specifically for textarea once this worked. Have reverted back to the tag and updated css. 
Also, textarea and input are both focusable elements with similar properties, so I don't think there should be much difference.

varaizad (approved):

PR #481, opened 2021-06-10
Option to highlight through keyboard

varaizad:
Adding action handlers on keyup and keydown events to allow the highlighting through keyboard. Also tjhe delete button wasn't focusable, so added tab-index to make it focusbale.

Will do the changes for aria-label in next commit.

Tested working as expected.


sugupta427 (approved):
> Have you checked the narrator behavior on highlighted text? Ideally the narrator should announce the selected content.

Yes the narrator will continue to work as it is working without these changes.

manu8996 (approved):

varaizad:
> Have you checked the narrator behavior on highlighted text? Ideally the narrator should announce the selected content.

Yes the narrator will continue to work as it is working without these changes.

PR #482, opened 2021-06-11
supporting 200% and 400% zoom scenarios

varaizad:
Earlier during 200% zoom, the panels were overlapping making the text hidden. Some options of the panel were getting out of the viewport and making then inaccessible.

Wrote the custom css for the components to support these scenarios and added a scrollbar to not let the content go out of the view port. 

Testing: With these changes I can support 200% zoom across all resolutions and 400% zoom across some resolutions.


ksboora (changes requested):
Pl don't merge these changes

PR #483, opened 2021-06-14
Adding aria-label to highlighter button

varaizad:
Adding aria-label to let users know to switch to caret mode if they want to highlight using keyboard. Changing the name of the variable to make it coherent with string, it is highlighter mode not article mode.

Testing: tested locally using narrator in edge and chrome, working as expected.


sugupta427 (approved):
> was it not possible to enter highlighter mode as soon as clipper launches, so that user do not have to anything special to highlight?


User settings in browser can't be changed using JS so we can't enable the caret mode while we switch to highlighter. Also in a discussion with PM, we felt that caret mode is already known to the accessibility users so I think this won't appear much of an overhead.

PR #484, opened 2021-06-17
Change PNG icons to SVGs

manu8996:
Issue: [Bug 3174405](https://office.visualstudio.com/OneNote/_workitems/edit/3174405): Multiple Icons were not visible in High Contrast White Mode.
Fix: 
- PNGs have been removed to introduce new font icons
- Introduced -ms-high-contrast media query to invert icon colours when Windows High Contrast mode is set. 
- A unit test that verifies that the default icon for augmentation mode is "article.png" has been removed because to verify the same for SVGs, we'll have to parse its xml differently to get icon src. Will take it up later. ([Bug 5197568](https://office.visualstudio.com/OneNote/_workitems/edit/5197568))

Test notes:
- Verified expected behaviour in Edge.

Caveats: 
- Chrome doesn't honour -ms-high-contrast media query and hence these changes won't be applicable for it. It has a 'High Contrast' extension though to provide accessibility support and is not used for Windows High Contrast a11y testing. ([Reference article](https://css-tricks.com/accessible-svgs-high-contrast-mode/))
- This PR fixes all the icons mentioned in the attached bug, but there are a few more that would be good to have but we are waiting on red-lines, some of which include the white spinner that appears momentarily in the first pop up when Clipper extension is invoked, tick mark icon in the 'Saved' prompt and the highlighter cursor. We'll resolve these in due time([Bug 5197565](https://office.visualstudio.com/OneNote/_workitems/edit/5197565)) but the PR should be sufficient to close the bug.


sugupta427:
can we provide the links of the new work items created in the PR description as well, so that it can be referred and tracked properly.  i.e. WI for unit test and WI for other remaining icons

sugupta427:
Have you tested the changes against all other high contrast modes?

manu8996:
> Have you tested the changes against all other high contrast modes?

Yes, verified it against all Windows High Contrast modes

manu8996:
> can we provide the links of the new work items created in the PR description as well, so that it can be referred and tracked properly. i.e. WI for unit test and WI for other remaining icons

Added Work Item links inline

sugupta427 (approved):

varaizad (approved):

PR #489, opened 2021-07-19
Add support for Region Selection using keyboard

susinh:
Issue: In region selection mode, we don't have a way to make the selection using the keyboard which is an accessibility issue.

Fix:
- Implemented a custom cursor for the region selection experience using a custom image for the cross-hair cursor
- Added handling to start region selection using Enter key stroke and make a selection using arrow keys.
- After making the selection, the selection can be completed using Enter key 
- The selection can be made using both mouse and keyboard and keyboard selection always get priority over keyboard
- Added global mouse movement handler which set the current position of the mouse
- Added global mouseOver handler to detect the initial state of the cursor
- In case the mouse is not within viewport, using the center of the window as the initial position for cursor
- Added keyDownhandler and keyUpHandler intercepting arrow and enter keystrokes to make a selection in the region selection canvas
- Removed the logic of initially setting focus to the back button in Region selecting panel and added code to set focus by default to selection region by updating the config attribute of for the RegionSelectionCanvas by updating the UpdateVisualElement callback for the dom element being passed to config attribute of mithril library.
- Updated the message in Region selection panel with the scrubbed strings providing better instruction to make the selection using mouse and keyboard
- Setting the role for selecion canvas as application as that looks like the most appropriate role when we are having custom handling with arrows and desktop app like behavior: More details here: https://www.w3.org/TR/wai-aria-1.1/#application
- Added logic to have hybrid cursor speed starting with 1 px and going up to 5px per key event. Resetting it back to 1px when a keyup for any arrow key is observed so that the user can make the selection precisely
- Added logic to support cursor movement diagonally by maintaining a keyDownDict which gets the arrow keys pressed at any particular moment and updating the mouse position based on this.
- Added handling to continue keyboard selection even when the actual mouse cursor is outside viewport 

Test Notes:
- Validated on Browsers Edge, Chrome and Firefox
- Validated that selection is possible using mouse
- Validated that selection using possible using keyboard
- Validated that while mouse selection if the user exists the viewport, the selection stops at last point inside the viewport as it was happening before when the cursor comes back 
- Validated that while keyboard selection is in progress user can continue with the selection using keyboard even if the cursor is outside the viewport.
- Validated that the Back and Exit buttons in the Region Selection panel are working correctly
- Validated that the correct alert is being called out by Narrator when entering Region mode
- Validated that the user is able to make a selection using the keyboard with a precision of 1 px if intended
- Validated the clipped content is being saved to the OneNote
- Validated that the user is able to select diagonally as well using multiple arrow keys
- Validated that the focus lands by default to the selection region and can be brought back using Tab
- Validated that it works fine with window resize


src/styles/clipper.less
```diff
@@ -175,7 +175,8 @@
 	margin: 0;
 	height: 100%;
 	width: 100%;
-	cursor: crosshair;
+	cursor: none;
+	position: absolute;
```
varaizad, line 179: Will it work fine if let's say window is resized? Did we test it there?

src/scripts/clipperUI/regionSelector.tsx
```diff
@@ -41,26 +51,29 @@ class RegionSelectorClass extends ComponentBase<RegionSelectorState, ClipperStat
 		this.resetState();
 
 		window.addEventListener("resize", this.resizeHandler);
+		window.addEventListener("mousemove", this.mouseMovementHandler);
+		window.addEventListener("mouseover", this.mouseOverHandler);
 	}
 
 	private onunload() {
 		window.removeEventListener("resize", this.resizeHandler);
+		window.removeEventListener("mousemove", this.mouseMovementHandler);
```
varaizad, line 60: Is it not needed to remove mouseover listener on unload?

src/styles/clipper.less
varaizad, line 179: Will it work fine if let's say window is resized? Did we test it there?

src/scripts/clipperUI/panels/regionSelectingPanel.tsx
```diff
@@ -8,25 +8,23 @@ class RegionSelectingPanelClass extends ComponentBase<{}, ClipperStateProp> {
 		this.props.clipperState.reset();
 	}
 
-	initiallySetFocusToBackButton(element: HTMLElement) {
-		element.focus();
-	}
```
varaizad, line 13: Are we accounting this somewhere else otherwise won't this cause a11y issue?

susinh: Yes. Back button is not the primary action on this screen anymore and the primary actionable item is the region selection canvas to which we are setting the focus by default.

src/scripts/clipperUI/regionSelector.tsx
susinh: I am removing it in the handler itself after we detect it once. But it might make sense to remove it in onunload as well just in case the mouse handler never came over the window.

src/styles/clipper.less
susinh: Validated that it works fine on window resize. We are already handling window resize event by adding a Window Resize handler.

varaizad (approved):

sugupta427 (approved):

PR #493, opened 2021-09-17
Removed feedback buttton

sugupta427:
Due to compliance reasons, feedback button can only be displayed after checking the age of the user, so for now removing feedback button.

Tested and verified that the feedback button does not show up on the sign page and main page for both work account and MSA accounts.

### DIFF
--- a/src/scripts/clipperUI/components/footer.tsx
+++ b/src/scripts/clipperUI/components/footer.tsx
@@ -75,10 +75,10 @@ class FooterClass extends ComponentBase<FooterState, FooterProps> {
 				style={Localization.getFontFamilyAsStyle(Localization.FontFamily.Regular)}>
 				<div className={Constants.Ids.footerButtonsContainer}>
 					<div className="footerButtonsLeft">
-						<a id={Constants.Ids.feedbackButton} role="button" {...this.enableInvoke({callback: this.handleFeedbackButton, tabIndex: 80})}>
+						{/* <a id={Constants.Ids.feedbackButton} role="button" {...this.enableInvoke({callback: this.handleFeedbackButton, tabIndex: 80})}>
 							<img id={Constants.Ids.feedbackImage} src={ExtensionUtils.getImageResourceUrl("feedback_smiley.svg")} aria-hidden="true"/>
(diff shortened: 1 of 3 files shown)


src/scripts/clipperUI/components/footer.tsx
```diff
@@ -75,10 +75,10 @@ class FooterClass extends ComponentBase<FooterState, FooterProps> {
 				style={Localization.getFontFamilyAsStyle(Localization.FontFamily.Regular)}>
 				<div className={Constants.Ids.footerButtonsContainer}>
 					<div className="footerButtonsLeft">
-						<a id={Constants.Ids.feedbackButton} role="button" {...this.enableInvoke({callback: this.handleFeedbackButton, tabIndex: 80})}>
+						{/* <a id={Constants.Ids.feedbackButton} role="button" {...this.enableInvoke({callback: this.handleFeedbackButton, tabIndex: 80})}>
 							<img id={Constants.Ids.feedbackImage} src={ExtensionUtils.getImageResourceUrl("feedback_smiley.svg")} aria-hidden="true"/>
 							<span id={Constants.Ids.feedbackLabel} class="buttonTextInHighContrast">{Localization.getLocalizedString("WebClipper.Action.Feedback") }</span>
-						</a>
+						</a> */}
 					</div>
```
varaizad, line 82: should we comment this div as well, otherwise it might cause some padding issues?

sugupta427: On commenting this div, there were some padding changes were seen, hence kept the div as it is.

manu8996 (approved):

varaizad (approved):

larsendaniel:
Could the check simply happen over at the destination? --> https://www.onenote.com/feedback

Does that page also need to check the user's age before they can ask the questions? (I assume so.) If so, then I propose no change here since it will be checked at the destination. :)

If they don't have to, then this change looks good to me. I also checked and the padding is correct with this exact change. 👍

sugupta427:
> Could the check simply happen over at the destination? --> https://www.onenote.com/feedback
> 
> Does that page also need to check the user's age before they can ask the questions? (I assume so.) If so, then I propose no change here since it will be checked at the destination. :)
> 
> If they don't have to, then this change looks good to me. I also checked and the padding is correct with this exact change. 👍

You are right! we are already working on it and hence have not merged this PR. I will abendon this PR as the removal of feedback button is not the solution to it.

PR #525, opened 2023-04-20
EUDB DataBoundary check in WebClipper Client by getting Email ID of OrgID user

gopal29392:
There are two changes done in this PR:
1. For an OrgID User, the email address info was not present in the UserInfoData returned by the Service. The email address is filled by decoding the token using jwt-decode and assigning the upn value of the decoded value to the email address.
2. Later on,for a valid user information the corresponding telemetry region is filled as the dataBoundary by using the function getDataBoundary().
3. The getDataBoundary uses the federation provider url https://odc.officeapps.live.com/odc/v2.1/federationprovider with a query parameter being domain. This query param value is set with domainValue extracted from the user's email address. The telemetry region value is the Data boundary that is needed from response to the url.
4. For EUDB, it will be provided with EMEA and for others it provided with GLOBAL as telemetry regions.

Tested locally to verify the email address extraction from the token and telemetry region 
MicrosoftTeams-image (1)
MicrosoftTeams-image

### DIFF
--- a/src/scripts/extensions/authenticationHelper.ts
+++ b/src/scripts/extensions/authenticationHelper.ts
@@ -15,7 +15,9 @@ import {ObjectUtils} from "../objectUtils";
 import {ResponsePackage} from "../responsePackage";
 import {StringUtils} from "../stringUtils";
 import {UserInfoData} from "../userInfo";
-import {UrlUtils} from "../urlUtils";
+import { UrlUtils } from "../urlUtils";
+import { DataBoundary } from "./DataBoundary";
+import * as fetch from "node-fetch";
 
 declare var browser;
 
@@ -58,7 +60,7 @@ export class AuthenticationHelper {
 
 			let getInfoEvent: Log.Event.PromiseEvent = new Log.Event.PromiseEvent(Log.Event.Label.GetExistingUserInformation);
 			getInfoEvent.setCustomProperty(Log.PropertyName.Custom.UserInformationStored, !!storedUserInformation);
-			this.clipperData.getFreshValue(ClipperStorageKeys.userInformation, getUserInformationFunction, updateInterval).then((response: TimeStampedData) => {
+			this.clipperData.getFreshValue(ClipperStorageKeys.userInformation, getUserInformationFunction, updateInterval).then(async (response: TimeStampedData) => {
 				let isValidUser = this.isValidUserInformation(response.data);
 				getInfoEvent.setCustomProperty(Log.PropertyName.Custom.FreshUserInfoAvailable, isValidUser);
 
@@ -68,6 +70,9 @@ export class AuthenticationHelper {
 				getInfoEvent.setCustomProperty(Log.PropertyName.Custom.UserUpdateReason, UpdateReason[updateReason]);
 
 				if (isValidUser) {
+					let userDataBoundary: string = await this.getUserDataBoundary(response.data);
+					getInfoEvent.setCustomProperty(Log.PropertyName.Custom.DataBoundary, userDataBoundary);
+					response.data.dataBoundary = userDataBoundary;
 					this.user.set({ user: response.data, lastUpdated: response.lastUpdated, updateReason: updateReason, writeableCookies: writeableCookies });
 				} else {
 					this.user.set({ updateReason: updateReason, writeableCookies: writeableCookies });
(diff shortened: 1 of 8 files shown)


src/scripts/extensions/authenticationHelper.ts
```diff
@@ -139,4 +143,43 @@ export class AuthenticationHelper {
 		// Note that we are returning true by default to ensure the N-1 scenario.
 		return userInfo.cookieInRequest !== undefined ? userInfo.cookieInRequest : true;
 	}
+
+	/**
+	 * fetch the user data bounday from the emailAddress
+	 * @param userInfo 
+	 * @returns user data boudary
+	 */
+	private async getUserDataBoundary(userInfo: UserInfoData): Promise<string | undefined> {
```
varaizad: Should this function be taken out as a util to some EUDB utilities file? Since it seems like it is not using any class members, so either it should be static or better take it out from this authenticationhelper class

src/scripts/extensions/authenticationHelper.ts
```diff
@@ -139,4 +143,43 @@ export class AuthenticationHelper {
...
+			if (!userInfo) {
+				return undefined;
+			}
+			if (userInfo.authType === AuthType[AuthType.Msa]) {
+				return DataBoundary[DataBoundary.GLOBAL];
+			}
+			let domainValue;
+			if (!userInfo.emailAddress) {
+				return undefined;
+			} else {
```
varaizad: Else not required since we are returning from if.

src/scripts/http/cachedHttp.ts
```diff
@@ -104,6 +106,19 @@ export class CachedHttp {
 			return undefined;
 		}
 
+		const userAuthType: string = valueAsJson.authType;
```
varaizad, line 109: So basically this piece of code will only be called when there is a need for a fresh value right? code here https://github.com/OneNoteDev/WebClipper/blob/dc7c9feafcbf4168712165f210afa43ae6908a8d/src/scripts/http/cachedHttp.ts#L28, which means there can still be no userinfo till the time cache is not expired. Is it fine or should we add the upn during boot also?

src/scripts/extensions/authenticationHelper.ts
```diff
@@ -139,4 +143,43 @@ export class AuthenticationHelper {
...
+				domainValue = userInfo.emailAddress.substring(
+					userInfo.emailAddress.indexOf("@") + 1
+				);
+			}
+			const urlDataBoundaryDomain: string = UrlUtils.addUrlQueryValue(Constants.Urls.userDataBoundaryDomain, Constants.Urls.QueryParams.domain, domainValue);
+			let response = await fetch(urlDataBoundaryDomain, {
```
varaizad: What would this API return in case of bad domain  value? Will it route to EU regions for it to be most compliant?

src/scripts/extensions/authenticationHelper.ts
gopal29392: Added a UserDataBoundaryHelper.ts file and extracted this logic to that file

src/scripts/extensions/authenticationHelper.ts
gopal29392: Resolved this

src/scripts/extensions/authenticationHelper.ts
gopal29392: In Case of bad domain value there wont be any telemetry region field in the response to this API. So used UNKNOWN as telemetry region  and this would confirm to EU region as per service changes

src/scripts/http/cachedHttp.ts
gopal29392: I see that whenever we retrieve user info with a new access token then it is necessary to decode the token and add that info to userinfo. For the cached token it does not make much sense to do that and eventually freshtoken will be called for the cached one as well once expired. So it is not necessary to handle for that.
Let me know if you feel otherwise

src/scripts/extensions/userDataBoundaryHelper.ts
```diff
@@ -0,0 +1,65 @@
+import { Clipper } from "../clipperUI/frontEndGlobals";
+import { Constants } from "../constants";
+import { HttpWithRetries } from "../http/HttpWithRetries";
+import { UrlUtils } from "../urlUtils";
+import { UserInfoData, AuthType } from "../userInfo";
+import { DataBoundary } from "./DataBoundary";
+
+export class UserDataBoundaryHelper {
```
varaizad, line 8: why do we want to make a class for it and not just utilities functions? it doesn't have any class members, so IMO it is fine to just add the functions.

gopal29392: Used class implementation for handling any future extensions to the class to add any more functions.

varaizad (approved):

PR #526, opened 2023-04-24
Fixing Header title for Selection Item

gopal29392:
Bug : In webclipper pane for both Article and Selection options, only Article title is appearing at header.
Problem : Article option is previewed with the help of AugmentationPreview class and Selection Option is previewed with the help of selectionPreview class. Both of these options are derived from the common base class EditorPreviewComponentBase.  This base class is used to set the header using getHeader function for both of them. The getHeader function uses PreviewViewerAugmentationHeader class to display all things related to Header for both of these options.  The PreviewViewerAugmentationHeader class has getTitle which is used to display header tile for both of these options. This function is common and uses the localization of Article for both the options. Hence in Selection option, Article is shown in header as title.

Solution : The Current mode of the Clipper State is set in getHeader function of EditorPreviewComponentBase class. This property is later used in getTitle of PreviewViewerAugmentationHeader class to differentiate between Article and Selection and use the respective Localization title.

Testing : tested the changes locally. Following are the screenshots before and after the changes.
Before : 
MicrosoftTeams-image (4)

After : 
MicrosoftTeams-image (2)

MicrosoftTeams-image (3)

### DIFF
--- a/src/scripts/clipperUI/components/previewViewer/editorPreviewComponentBase.tsx
+++ b/src/scripts/clipperUI/components/previewViewer/editorPreviewComponentBase.tsx
@@ -75,7 +75,8 @@ export abstract class EditorPreviewComponentBase<TState extends EditorPreviewSta
 			changeFontFamily={this.changeFontFamily.bind(this)}
 			changeFontSize={this.changeFontSize.bind(this)}
 			serif={this.props.clipperState.previewGlobalInfo.serif}
-			textHighlighterEnabled={this.props.clipperState.previewGlobalInfo.highlighterEnabled} />;
+			textHighlighterEnabled={this.props.clipperState.previewGlobalInfo.highlighterEnabled}
+			currentMode={this.props.clipperState.currentMode} />;
 	}
 
 	// Override
--- a/src/scripts/clipperUI/components/previewViewer/previewViewerAugmentationHeader.tsx
+++ b/src/scripts/clipperUI/components/previewViewer/previewViewerAugmentationHeader.tsx
@@ -1,6 +1,8 @@
-﻿import {Constants} from "../../../constants";
+﻿import { SmartValue } from "../../../communicator/smartValue";
+import {Constants} from "../../../constants";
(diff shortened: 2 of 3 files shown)


src/scripts/clipperUI/components/previewViewer/previewViewerAugmentationHeader.tsx
ksboora, line 1: what would happen for other modes like region?

gopal29392: The other modes have their own header implementations . For example Region mode has its header implemented in PreviewViewerRegionHeader class. Only the Article and Selection modes are tied to a common base header class PreviewViewerAugmentationHeader

src/scripts/clipperUI/components/previewViewer/editorPreviewComponentBase.tsx
```diff
@@ -75,7 +75,8 @@ export abstract class EditorPreviewComponentBase<TState extends EditorPreviewSta
 			changeFontFamily={this.changeFontFamily.bind(this)}
 			changeFontSize={this.changeFontSize.bind(this)}
 			serif={this.props.clipperState.previewGlobalInfo.serif}
-			textHighlighterEnabled={this.props.clipperState.previewGlobalInfo.highlighterEnabled} />;
+			textHighlighterEnabled={this.props.clipperState.previewGlobalInfo.highlighterEnabled}
+			currentMode={this.props.clipperState.currentMode} />;
```
varaizad, line 79: where is currentMode being set? Are there two different class for these modes, since in mock test I see only augmentation.

gopal29392: These stated are set when any of the options is selected in ModeButtonSelectorClass class . Yes in mock test currently Augmentation mode is used for testing (anyways augmentation and select options are using same base class for header)

varaizad (approved):

ksboora (approved):

PR #528, opened 2023-05-29
Reinitializing Aria Logger with the DataBoundary Pipeline config

gopal29392:
Currently the default config pipeline for the Aria logging is done through EU route. This way we usually log to EU route before signing in. Once we sign in and get to know about the user data boundary, if the user data boundary is different from EUDB , Reinitialization of Aria Logger on LogManager is done. The changes related to reinit are part of WebClipper internal repo.

### DIFF
--- a/src/scripts/extensions/authenticationHelper.ts
+++ b/src/scripts/extensions/authenticationHelper.ts
@@ -75,7 +75,7 @@ export class AuthenticationHelper {
 					// The default logging has been configured to EU Pipeline. Once we find the
 					// userdataboundary and if it is different from EUDB , reinit the logger with WW Pipeline
 					if (userDataBoundary === DataBoundary[DataBoundary.GLOBAL] || userDataBoundary === DataBoundary[DataBoundary.PUBLIC]) {
-						LogManager.reInitLoggerForDataBoundaryChange();
+						LogManager.reInitLoggerForDataBoundaryChange(userDataBoundary);
 					}
 					getInfoEvent.setCustomProperty(Log.PropertyName.Custom.DataBoundary, userDataBoundary);
 					response.data.dataBoundary = userDataBoundary;
--- a/src/scripts/extensions/authenticationHelper.ts
+++ b/src/scripts/extensions/authenticationHelper.ts
@@ -17,6 +17,7 @@ import {StringUtils} from "../stringUtils";
 import {UserInfoData} from "../userInfo";
 import { UrlUtils } from "../urlUtils";
 import { UserDataBoundaryHelper } from "./userDataBoundaryHelper";
+import { DataBoundary } from "./DataBoundary";
 
 declare var browser;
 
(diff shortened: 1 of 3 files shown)


src/scripts/extensions/authenticationHelper.ts
```diff
@@ -71,6 +72,11 @@ export class AuthenticationHelper {
 				if (isValidUser) {
 					const dataBoundaryHelper = new UserDataBoundaryHelper();
 					let userDataBoundary: string = await dataBoundaryHelper.getUserDataBoundary(response.data);
+					// The default logging has been configured to EU Pipeline. Once we find the
+					// userdataboundary and if it is different from EUDB , reinit the logger with WW Pipeline
+					if (userDataBoundary === DataBoundary[DataBoundary.GLOBAL] || userDataBoundary === DataBoundary[DataBoundary.PUBLIC]) {
+						LogManager.reInitLoggerForDataBoundaryChange();
```
varaizad: Does reinitializing mean we will log all the logs in buffer to the new region or partly to EU and partly to WW if data boundary in non-EU?

src/scripts/extensions/authenticationHelper.ts
```diff
@@ -71,6 +72,11 @@ export class AuthenticationHelper {
 				if (isValidUser) {
 					const dataBoundaryHelper = new UserDataBoundaryHelper();
 					let userDataBoundary: string = await dataBoundaryHelper.getUserDataBoundary(response.data);
+					// The default logging has been configured to EU Pipeline. Once we find the
+					// userdataboundary and if it is different from EUDB , reinit the logger with WW Pipeline
+					if (userDataBoundary === DataBoundary[DataBoundary.GLOBAL] || userDataBoundary === DataBoundary[DataBoundary.PUBLIC]) {
```
varaizad, line 77: Also what is the flushing logic for this? Does a network request goes for every log line or is it batched and then sent?

src/scripts/extensions/authenticationHelper.ts
gopal29392: The logs that are captured till this point are already logged to EU region. Once we determine the region, if it is Non EU, then we log to WW else it will be logged to EU.
This way we also try to solve for the case where in before sign in happens , the default logging is going to happen in EU only

src/scripts/extensions/authenticationHelper.ts
gopal29392: Flushing logic is handled in batches by Aria-Web-Telemetry logger library which captures all the logging happening into its worker thread cache and flushes periodically.

src/scripts/extensions/authenticationHelper.ts
```diff
@@ -71,6 +72,11 @@ export class AuthenticationHelper {
...
+						LogManager.reInitLoggerForDataBoundaryChange(userDataBoundary);
```
varaizad, line 78: since this is implemented in web clipper internal, how do we ensure the implementation of this function is also available along with this deployment of clipper? Just wanted to know what happens when the implementation is not available, since build will not fail because we have added this in types (.d.ts file), but will it result in run time crash?

src/scripts/extensions/authenticationHelper.ts
varaizad: So with your first statement if the user belong to WW, then the logs before signin goes to EU region and after sigin they go to WW region? Not sure how we will correlate and track the funnel in this case

gopal29392: Yes, the logs go to EU region before signing in and after signing in, depending upon boundary they go to respective region. To correlate and track the funnel , i used the DeviceInfo_Id which is common for the session

src/scripts/extensions/authenticationHelper.ts
gopal29392: Added the function definition locally as well which does not do any reinit as that can happen only in respective tenant environments.

varaizad (approved):

PR #541, opened 2024-09-20
Web Clipper Migration to Manifest v3

ArpanKundu97:
This PR is to migrate Web Clipper to Manifest v3.

Changes included:
1. Move DOM calls to an offscreen document
2. Move localStorage calls to an offscreen document

Reference: https://developer.chrome.com/docs/extensions/develop/migrate

### DIFF
--- a/src/scripts/domParsers/domUtils.ts
+++ b/src/scripts/domParsers/domUtils.ts
@@ -487,7 +487,7 @@ export class DomUtils {
 			location = document.location;
 		}
 
-		let bases: NodeList = doc.head.getElementsByTagName(DomUtils.tags.base);
+		let bases: NodeList = doc.head.querySelectorAll(DomUtils.tags.base);
 		if (bases.length === 0) {
 			let baseUrl = location.href.split("#")[0].split("?")[0];
 			baseUrl = baseUrl.substr(0, baseUrl.lastIndexOf("/") + 1);
--- a/src/scripts/domParsers/domUtils.ts
+++ b/src/scripts/domParsers/domUtils.ts
@@ -205,7 +205,7 @@ export class DomUtils {
 	 * removes any base64 encoded binaries defined in any <style> tags.
 	 */
 	public static removeStylesWithBase64EncodedBinaries(doc: Document): void {
-		DomUtils.domReplacer(doc, "style", (node: HTMLElement) => {
+		DomUtils.domReplacer(doc, "style", async (node: HTMLElement) => {
 			return node.innerHTML.indexOf("data:application") !== -1 ? undefined : node;
 		});
 	}
@@ -217,27 +217,27 @@ export class DomUtils {
 		let tagsToTurnIntoDiv = [DomUtils.tags.main, DomUtils.tags.article, DomUtils.tags.figure, DomUtils.tags.header, DomUtils.tags.center];
 
 		// ... and for everything else, we replace them with an equivalent, preserving the inner HTML
-		DomUtils.domReplacer(doc, tagsToTurnIntoDiv.join(), (node: HTMLElement) => {
+		DomUtils.domReplacer(doc, tagsToTurnIntoDiv.join(), async (node: HTMLElement) => {
 			let div = document.createElement("div");
 			div.innerHTML = DomUtils.cleanHtml(node.innerHTML);
 			return div;
 		});
 	}
 
-	public static domReplacer(doc: Document, querySelector: string, getReplacement: (oldNode: Node, index: number) => Node = () => undefined) {
+	public static domReplacer(doc: Document, querySelector: string, getReplacement: (oldNode: Node, index: number) => Promise<Node> = () => undefined) {
 		let nodes: NodeList = doc.querySelectorAll(querySelector);
 
 		for (let i = 0; i < nodes.length; i++) {
 			let oldNode: Node = nodes[i];
 
 			try {
-				let newNode = getReplacement(oldNode, i);
-
-				if (!newNode) {
-					oldNode.parentNode.removeChild(oldNode);
-				} else if (oldNode !== newNode) {
-					oldNode.parentNode.replaceChild(newNode, oldNode);
-				}
+				getReplacement(oldNode, i).then((newNode) => {
+					if (!newNode) {
+						oldNode.parentNode.removeChild(oldNode);
+					} else if (oldNode !== newNode) {
+						oldNode.parentNode.replaceChild(newNode, oldNode);
+					}
+				});
 			} catch (e) {
 				// There are some cases (like dirty canvases) where running replace will throw an error.
 				// We catch it, thus leaving the original.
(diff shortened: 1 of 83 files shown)


src/scripts/storage/localStorage.ts
ipdutta, line 1: The calls to window.localStorage in this file needs to be replaced with offscreen document calls

src/scripts/logging/loggerDecorator.ts
```diff
@@ -128,7 +128,7 @@ export abstract class LoggerDecorator extends SessionLogger {
 	}
 
 	protected handleSetContext(key: Log.Context.Custom, value: string | number | boolean): void {
-		this.outputSetContext(key, value);
+		// this.outputSetContext(key, value);
```
ipdutta: This change needs to be reverted

src/scripts/logging/logHelpers.ts
```diff
@@ -152,11 +152,11 @@ export module LogHelpers {
 	}
 
 	export function isConsoleOutputEnabled(): boolean {
-		try {
+		/* try {
```
ipdutta: Needs to be replaced with a direct call to offscreen document

src/scripts/saveToOneNote/oneNoteApiWithRetries.ts
```diff
@@ -1,84 +1,221 @@
...
+				 * Otherwise, it is of type OneNoteApi.RequestError.
+				 */
+				if ("parsedResponse" in response) {
+					resolve(response);
+				} else {
+					reject(response);
+				}
+			}).catch((error) => {
+				reject(error);
+			});
```
ipdutta, line 30: For my understanding - why do we need to have this additional check now since earlier we were directly returning.

src/scripts/extensions/extensionBase.ts
```diff
@@ -56,30 +58,34 @@ export abstract class ExtensionBase<TWorker extends ExtensionWorkerBase<TTab, TT
...
-		if (!clipperId) {
-			// New install
-			clipperFirstRun = true;
-			clipperId = ExtensionBase.generateClipperId();
-			this.clipperData.setValue(ClipperStorageKeys.clipperId, clipperId);
-
-			// Ensure fresh installs don't trigger thats What's New experience
-			this.updateLastSeenVersionInStorageToCurrent();
-		}
+		this.offscreenDocumentResultProcessed = sendToOffscreenDocument("get-from-local-storage", {
```
ipdutta: Needs to be reverted once the storage and offscreen document changes are merged

src/scripts/extensions/webExtensionBase/webExtension.ts
```diff
@@ -40,7 +41,10 @@ export class WebExtension extends ExtensionBase<WebExtensionWorker, W3CTab, numb
 		this.injectUrls = injectUrls;
 
 		this.registerBrowserButton();
-		this.registerContextMenuItems();
+
+		this.offscreenDocumentResultProcessed.then(() => {
```
ipdutta: Just for my understanding - what does this.offscreenDocumentResultProcessed do and why are we waiting for it to call some of the functions in this file which were called normally before?

src/scripts/extensions/chrome/manifest.json
ipdutta, line 1: Although we are not deploying to edge should we make the parallel manifest and offscreen document changes for edge as well?

src/scripts/extensions/chrome/chromeOffscreen.ts
ipdutta, line 1: Remaining local storage functionality needs to be added

src/scripts/domParsers/domUtils.ts
```diff
@@ -217,27 +217,27 @@ export class DomUtils {
 		let tagsToTurnIntoDiv = [DomUtils.tags.main, DomUtils.tags.article, DomUtils.tags.figure, DomUtils.tags.header, DomUtils.tags.center];
 
 		// ... and for everything else, we replace them with an equivalent, preserving the inner HTML
-		DomUtils.domReplacer(doc, tagsToTurnIntoDiv.join(), (node: HTMLElement) => {
+		DomUtils.domReplacer(doc, tagsToTurnIntoDiv.join(), async (node: HTMLElement) => {
 			let div = document.createElement("div");
```
ipdutta: Does document.createElement work here and not need to be moved to offscreen document?

src/scripts/extensions/extensionBase.ts
```diff
@@ -56,7 +57,8 @@ export abstract class ExtensionBase<TWorker extends ExtensionWorkerBase<TTab, TT
 
 		let clipperFirstRun = false;
 
-		this.clipperData.getValue(ClipperStorageKeys.clipperId).then((clipperId) => {
+		this.offscreenDocumentResultProcessed = this.clipperData.getValue(ClipperStorageKeys.clipperId).then((clipperId) => {
+			console.log("clipperId from storage:", clipperId);
```
ipdutta: Should we remove this logging once testing is done

src/scripts/logging/loggerDecorator.ts
ArpanKundu97: Already reverted in latest iteration.

src/scripts/logging/logHelpers.ts
ArpanKundu97: Done. Also, verified that console logging works by setting the "enable_console_logging" local storage entry in the offscreen document to true.

image

src/scripts/saveToOneNote/oneNoteApiWithRetries.ts
ArpanKundu97: The interface `IOneNoteApi` has been replicated in responsePackage.ts from the OneNoteApi library so as to modify the return type of each function from a `Promise<ResponsePackage<any>>` to a `Promise<ResponsePackage<any> | OneNoteApi.RequestError>` to match the return types of the functions in the OneNoteApi.OneNoteApi class.

As a result, in this particular file, response could earlier only be a `ResponsePackage<any>` but can now also be `OneNoteApi.RequestError` and hence an additional else block to handle the same.

src/scripts/extensions/webExtensionBase/webExtension.ts
ArpanKundu97: The completion of this variable indicates that the clipper id has been successfully processed and the clientInfo has been successfully set, without which calling `registerContextMenuItems` will fail.

The code for this variable can be found in the base class `ExtensionBase`.

src/scripts/extensions/chrome/manifest.json
ArpanKundu97: The plan is to take care of this in a separate iteration since making the changes for edge and testing them out in edge might take some time.

src/scripts/domParsers/domUtils.ts
ArpanKundu97: I see that this is a part of the content scripts, i.e. `clipper.js` and not the service worker, i.e. `chromeExtension.js` and should hence continue working as expected.

ipdutta (approved):

src/scripts/extensions/extensionBase.ts
ArpanKundu97: This has been removed in the latest upload.

PR #542, opened 2024-10-04
Migrate XMLHttpRequest to fetch

ArpanKundu97:
As part of migrating OneNote Web Clipper from Manifest V2 to V3, we need to replace XMLHttpRequest() with global fetch() since XMLHttpRequest() can't be called from a service worker. This PR is to take care of the same.

Reference: https://developer.chrome.com/docs/extensions/develop/migrate/to-service-workers

NOTE: The typescript version has been updated from 2.1.6 to 2.3.4 in the package.json file to support the fetch() method.

### DIFF
--- a/src/scripts/contentCapture/augmentationHelper.ts
+++ b/src/scripts/contentCapture/augmentationHelper.ts
@@ -14,7 +14,7 @@ import {HttpWithRetries} from "../http/httpWithRetries";
 import * as Log from "../logging/log";
 
 import {CaptureFailureInfo} from "./captureFailureInfo";
-import { ResponsePackage } from "../responsePackage";
+import { ErrorUtils, ResponsePackage } from "../responsePackage";
 
 export enum AugmentationModel {
 	None,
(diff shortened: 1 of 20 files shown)


src/scripts/contentCapture/augmentationHelper.ts
```diff
@@ -102,30 +103,34 @@ export class AugmentationHelper {
...
+				headers[Constants.HeaderValues.userSessionIdKey] = sessionId;
+
+				HttpWithRetries.post(augmentationApiUrl, pageContent, headers).then((response: Response) => {
+					response.text().then((responseText: string) => {
+						let parsedResponse: any;
+						try {
+							parsedResponse = JSON.parse(responseText);
+						} catch (e) {
+							Clipper.logger.logJsonParseUnexpected(responseText);
+							reject(OneNoteApi.RequestErrorType.UNABLE_TO_PARSE_RESPONSE);
```
ipdutta: Will it be valuable while debugging failure conditions if we pass the response object in reject?
NOTE:- This applies to multiple files where reject has been used

ipdutta (approved):

src/scripts/contentCapture/augmentationHelper.ts
ArpanKundu97: Sure. However, we cannot directly use `OneNoteApi.ErrorUtils.createRequestErrorObject` since this function accepts a variable of type `XMLHttpRequest` as its first argument. Hence, I have now created a copy of this function within the WebClipper repo itself such that it accepts a `Response` as its first argument, and have used this function instead.

PR #551, opened 2024-11-05
Changes to the clip PDF flow and notifications

ipdutta:
Changes included -
1) PDF XMLHttpRequest to fetch
2) window.location.href to self.location.href
3) alert to notifications in background script

### DIFF
--- a/src/scripts/contentCapture/pdfScreenshotHelper.ts
+++ b/src/scripts/contentCapture/pdfScreenshotHelper.ts
@@ -28,7 +28,6 @@ export class PdfScreenshotHelper {
 
 	public static getPdfData(url: string): Promise<PdfScreenshotResult> {
 		return new Promise<PdfScreenshotResult>((resolve, reject) => {
-			debugger;
 			let getBinaryEvent = new Log.Event.PromiseEvent(Log.Event.Label.GetBinaryRequest);
 
 			let errorCallback = (failureInfo: Promise<OneNoteApi.RequestError>) => {
@@ -64,7 +63,7 @@ export class PdfScreenshotHelper {
 			});
 			setTimeout(() => {
 				reject(OneNoteApi.RequestErrorType.REQUEST_TIMED_OUT)
-			}, 1000);
+			}, 60000);
 		});
 	}
 
--- a/src/scripts/contentCapture/pdfScreenshotHelper.ts
+++ b/src/scripts/contentCapture/pdfScreenshotHelper.ts
@@ -13,6 +13,7 @@ import {CaptureFailureInfo} from "./captureFailureInfo";
 import {PdfDocument} from "./pdfDocument";
 import {PdfJsDocument} from "./pdfJsDocument";
 import {ViewportDimensions} from "./viewportDimensions";
+import { ErrorUtils } from "../responsePackage";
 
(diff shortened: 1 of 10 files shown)


src/scripts/extensions/injectHelper.ts
```diff
@@ -1,12 +1,19 @@
...
 	private static uninjectableUrlRegexes: RegExp[] = [
 		/^about:/
 	];
 
 	public static alertUserOfUnclippablePage() {
-		alert(Localization.getLocalizedString("WebClipper.Error.CannotClipPage"));
+		WebExtension.browser.notifications.create({
+			type: 'basic',
+			iconUrl: '/icons/icon-48.png',
+			title: 'Cannot Clip Page',
```
ArpanKundu97: Shall we localize the string for title as well?

src/scripts/contentCapture/pdfScreenshotHelper.ts
```diff
@@ -29,42 +30,40 @@ export class PdfScreenshotHelper {
...
 			};
 
-			request.onload = () => {
-				if (request.status === 200 && request.response) {
-					let arrayBuffer = request.response;
 
+			fetch(url)
+			.then(response => {
+				if (!response.ok) {
+					errorCallback(ErrorUtils.createRequestErrorObject(response, OneNoteApi.RequestErrorType.UNEXPECTED_RESPONSE_STATUS));
```
ArpanKundu97, line 47: `createRequestErrorObject` now returns a Promise. Should we add a `then` block here?

src/scripts/contentCapture/pdfScreenshotHelper.ts
```diff
@@ -29,42 +30,40 @@ export class PdfScreenshotHelper {
...
-			};
-
-			request.send();
+					PdfScreenshotHelper.getPdfScreenshotResult(new Uint8Array(arrayBuffer))
+						.then(pdfScreenshotResult => {
+							pdfScreenshotResult.byteLength = arrayBuffer.byteLength;
+							resolve(pdfScreenshotResult);
+						})
+				})
+			.catch(() => {
```
ArpanKundu97, line 61: Should we create a request error object here as well?

src/scripts/extensions/injectHelper.ts
ipdutta: Done

src/scripts/contentCapture/pdfScreenshotHelper.ts
ipdutta: My rationale behind this is that we pass the Promise object directly into the errorCallback function and have a then block inside it to resolve the promise. Let me know if you see a problem with this approach.

src/scripts/contentCapture/pdfScreenshotHelper.ts
ipdutta: The response object will not be available in the catch block hence not able to create a request error object here

src/scripts/contentCapture/pdfScreenshotHelper.ts
ArpanKundu97: Got it, thanks!

ArpanKundu97 (approved):

PR #554, opened 2024-11-14
Frame Url Injection Changes

ArpanKundu97:
As part of this PR, the frame url injection logic has been modified so as to avoid injecting the frame url from `webExtensionWorker.ts` which will eventually be required as part of the effort of migrating OneNote Web Clipper from Manifest V2 to V3.

### DIFF
--- a/src/scripts/extensions/bookmarklet/bookmarkletInject.ts
+++ b/src/scripts/extensions/bookmarklet/bookmarkletInject.ts
@@ -1,4 +1,3 @@
-declare var frameUrl: string;
 declare var bookmarkletRoot: string;
 
 import {IFrameMessageHandler} from "../../communicator/iframeMessageHandler";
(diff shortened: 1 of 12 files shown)


src/scripts/extensions/bookmarklet/bookmarkletInject.ts
```diff
@@ -1,4 +1,3 @@
-declare var frameUrl: string;
```
ipdutta, line 1: Do we know the implications of removing frameUrl from this file? Bookmarklet is another client type and it would be good to test the conditions when this gets invoked.

ArpanKundu97: Hi @ipdutta, bookmarklet has been deprecated. This is also evident from the fact that we aren't setting `webClipperInjectUrl` to "bookmarkletInject.js" anywhere. Hence, we can be sure that these changes won't have any implications.

ipdutta (approved):

PR #557, opened 2024-11-20
Changes to Auth Screen placement and Sign Out flow

ArpanKundu97:
This PR is to make changes to the way in which the authentication screen is displayed and to the sign out flow so that they work with Manifest V3.

### DIFF
--- a/src/scripts/extensions/webExtensionBase/webExtensionWorker.ts
+++ b/src/scripts/extensions/webExtensionBase/webExtensionWorker.ts
@@ -1,5 +1,4 @@
 import {AuthType} from "../../userInfo";
-import {BrowserUtils} from "../../browserUtils";
 import {ClientInfo} from "../../clientInfo";
 import {ClientType} from "../../clientType";
 import {ClipperUrls} from "../../clipperUrls";
@@ -71,7 +70,7 @@ export class WebExtensionWorker extends ExtensionWorkerBase<W3CTab, number> {
 	protected doSignOutAction(authType: AuthType) {
 		let usidQueryParamValue = this.getUserSessionIdQueryParamValue();
 		let signOutUrl = ClipperUrls.generateSignOutUrl(this.clientInfo.get().clipperId, usidQueryParamValue, AuthType[authType]);
-		BrowserUtils.appendHiddenIframeToDocument(signOutUrl);
+		fetch(signOutUrl);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


src/scripts/extensions/webExtensionBase/webExtensionWorker.ts
```diff
@@ -71,7 +70,7 @@ export class WebExtensionWorker extends ExtensionWorkerBase<W3CTab, number> {
 	protected doSignOutAction(authType: AuthType) {
 		let usidQueryParamValue = this.getUserSessionIdQueryParamValue();
 		let signOutUrl = ClipperUrls.generateSignOutUrl(this.clientInfo.get().clipperId, usidQueryParamValue, AuthType[authType]);
-		BrowserUtils.appendHiddenIframeToDocument(signOutUrl);
+		fetch(signOutUrl);
```
ipdutta, line 73: For my information - why are we doing a fetch call here? I believe the previous code fragment was adding a hidden iframe

src/scripts/extensions/webExtensionBase/webExtensionWorker.ts
```diff
@@ -195,78 +194,80 @@ export class WebExtensionWorker extends ExtensionWorkerBase<W3CTab, number> {
...
-					windowOptions.focused = true;
-				}
-
-				WebExtension.browser.windows.create(windowOptions, (newWindow: Window) => {
-					let redirectOccurred = false;
-					let errorObject;
-					let correlationId: string;
+			WebExtension.browser.windows.getCurrent((currentWindow: Window) => {
+				let leftPosition: number = currentWindow.left + Math.round((currentWindow.width - popupWidth) / 2);
+				let topPosition: number = currentWindow.top + Math.round((currentWindow.height - popupHeight) / 2);
```
ipdutta, line 199: Looks like there is a slight change in the position calculation here .. were we able to test if the placement is right in both edge and chrome?

ArpanKundu97: Yes, I have tested that the placement is right in Chrome (with both MV2 and MV3) as well as Edge (with MV2).

src/scripts/extensions/webExtensionBase/webExtensionWorker.ts
ArpanKundu97: Yes, the previous code was adding a hidden iframe using DOM elements which won't be possible now. I explored using the same function which is used for sign in, i.e. `launchWebExtensionPopupAndWaitForClose`. However, it would result in the creation of an additional sign out screen, just as we get an additional screen while signing in, and we would no longer have a 1-click sign out experience. Directly fetching the sign out url instead resolves this.

ipdutta (approved):

PR #559, opened 2024-11-27
Updated all storage functions to async

ipdutta:
Updated the get value from storage functionality to async

### DIFF
--- a/src/scripts/storage/storage.ts
+++ b/src/scripts/storage/storage.ts
@@ -1,6 +1,6 @@
 export interface Storage {
-	getValue(key: string): string;
-	getValues(keys: string[]): {};
+	getValue(key: string): Promise<string>;
+	getValues(keys: string[]): Promise<{}>;
 	setValue(key: string, value: string): void;
 	removeKey(key: string): boolean;
 }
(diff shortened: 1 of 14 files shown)


src/scripts/storage/storage.ts
```diff
@@ -1,6 +1,6 @@
 export interface Storage {
-	getValue(key: string): string;
-	getValues(keys: string[]): {};
+	getValue(key: string): Promise<string>;
+	getValues(keys: string[]): Promise<{}>;
 	setValue(key: string, value: string): void;
 	removeKey(key: string): boolean;
```
ArpanKundu97, line 5: Do we not need to make `removeKey` async as well?

ipdutta: As per the current implementation of removeKey it returns true whether or not the remove functionality is successful in storage hence we do not need to make it async .. we can return true as an indication that the remove functionality has been called and the actual task can happen async.

ArpanKundu97: Makes sense. Also, I see that `removeKey` is only being used in `cachedHttp.ts` where the remaining code doesn't seem to be dependent on the key removal to be complete. Hence, keeping `removeKey` as is should be fine.

ArpanKundu97 (approved):

PR #562, opened 2024-12-04
Idle Auth Fix

ArpanKundu97:
**Issue**
A signed in user is shown the sign in panel again after a certain period of inactivity.

**Cause**
The user information is no longer present in the state variable once the service worker becomes inactive.

**Fix**
Create a new variable in local storage to indicate if a user is logged in. If this variable is set to true, and the state variable does not already contain the user information, then fetch the user information and store it in the state variable.

### DIFF
--- a/src/scripts/clipperUI/clipperStateUtilities.ts
+++ b/src/scripts/clipperUI/clipperStateUtilities.ts
@@ -14,10 +14,10 @@ import { OffscreenMessageTypes } from "../communicator/offscreenMessageTypes";
 import { ClipperStorageKeys } from "../storage/clipperStorageKeys";
 
 export module ClipperStateUtilities {
-	export function isUserLoggedIn(state: ClipperState): boolean {
+	export function isUserLoggedIn(state: ClipperState, refreshUserInfo?: boolean): boolean {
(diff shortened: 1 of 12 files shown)


src/scripts/clipperUI/clipperStateUtilities.ts
```diff
@@ -1,13 +1,81 @@
...
+								userInfoUpdateCb(freshInitialUser);
+							}
+							Clipper.getExtensionCommunicator().subscribeAcrossCommunicator(new SmartValue<UserInfo>(), Constants.SmartValueKeys.user, (updatedUser: UserInfo) => {
+								userInfoUpdateCb(updatedUser);
+							});
+						}
+					});
+				}
+			});
+			/**
```
ipdutta, line 72: Didn't face any issue while testing but as per this comment - is there a chance that for a logged in user we will momentarily see the sign in panel and then once the offscreen task is completed we'll see the logged in state?

ipdutta (approved):

src/scripts/clipperUI/clipperStateUtilities.ts
ArpanKundu97: I have tested sign in and sign out multiple times with the changes in this PR, and each time I was able to see the correct panel.

PR #565, opened 2024-12-16
Fixed tooltip rendering and upgraded version to 3.10.1

ipdutta:
Fixed continuous tooltip rendering by adding async await to the verification functions and updated version to 3.10.1

### DIFF
--- a/src/scripts/extensions/extensionBase.ts
+++ b/src/scripts/extensions/extensionBase.ts
@@ -244,11 +244,11 @@ export abstract class ExtensionBase<TWorker extends ExtensionWorkerBase<TTab, TT
 		let type = this.checkIfTabMatchesATooltipType(tab, tooltipTypes);
 
 		if (!type) {
-			return;
+			return Promise.reject(undefined);
 		}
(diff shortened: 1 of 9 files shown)


src/scripts/extensions/extensionBase.ts
```diff
@@ -240,14 +240,14 @@ export abstract class ExtensionBase<TWorker extends ExtensionWorkerBase<TTab, TT
 		this.updateClientInfoWithFlightInformation([]);
 	}
 
-	private shouldShowTooltip(tab: TTab, tooltipTypes: TooltipType[]): TooltipType {
+	private async shouldShowTooltip(tab: TTab, tooltipTypes: TooltipType[]): Promise<TooltipType> {
 		let type = this.checkIfTabMatchesATooltipType(tab, tooltipTypes);
 
 		if (!type) {
```
SiddharthSinghMOD, line 246: should we return a failed promise instead of plain return?

SiddharthSinghMOD (approved):

PR #568, opened 2024-12-17
Disable clipper button on unclippable pages

ipdutta:
Disables clipper on all pages matching the regex **chrome://***

### DIFF
--- a/src/scripts/extensions/webExtensionBase/webExtension.ts
+++ b/src/scripts/extensions/webExtensionBase/webExtension.ts
@@ -46,6 +46,7 @@ export class WebExtension extends ExtensionBase<WebExtensionWorker, W3CTab, numb
 			this.registerContextMenuItems();
 		});
 		this.registerInstallListener();
+		this.registerTabCreateOrUpdateListener();
 		this.registerTabRemoveListener();
 	}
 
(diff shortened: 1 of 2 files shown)


src/scripts/extensions/webExtensionBase/webExtension.ts
```diff
@@ -242,6 +243,19 @@ export class WebExtension extends ExtensionBase<WebExtensionWorker, W3CTab, numb
 		});
 	}
 
+	private registerTabCreateOrUpdateListener() {
+		WebExtension.browser.tabs.onCreated.addListener((tab: W3CTab) => {
+		if (tab.url.match("chrome://*")) {
```
ArpanKundu97: Shall we check the value of WebExtension.browser (chrome or edge) and decide the regex accordingly instead of hardcoding it to "chrome://*"?

src/scripts/extensions/webExtensionBase/webExtension.ts
```diff
@@ -242,6 +243,19 @@ export class WebExtension extends ExtensionBase<WebExtensionWorker, W3CTab, numb
...
+			WebExtension.browser.action.disable(tab.id);
+		}
+		});
+		WebExtension.browser.tabs.onUpdated.addListener((tabId: number, changeInfo: any, tab: W3CTab) => {
```
SiddharthSinghMOD, line 258: do we need to listen to tab switch as well?

src/scripts/extensions/webExtensionBase/webExtension.ts
ipdutta: Made the code generic to handle multiple browsers

src/scripts/extensions/webExtensionBase/webExtension.ts
ipdutta: No the listeners get created for individual tabs so a tab switch would have separate listener

ArpanKundu97 (approved):

PR #572, opened 2024-12-26
Update default cluster from EU to ROW

ArpanKundu97:
The default cluster is being updated from EU to ROW as part of https://github.com/OneNoteDev/WebClipper_Internal/pull/67.

This PR is to take care of the corresponding changes in the WebClipper repo. Also, made sure to pass `Enlightened-Hrd-Client: 1` header to Email HRD call, along with other necessary headers.


src/scripts/extensions/userDataBoundaryHelper.ts
```diff
@@ -39,7 +40,21 @@ export class UserDataBoundaryHelper {
...
+				/**
+				 * TODO: Update the officeApplicationNumber to the correct value
+				 * once OneNote Web Clipper is onboarded to HRD.
+				 */
+				"X-Office-Application": undefined,
+				"Enlightened-Hrd-Client": "1", // To turn on the EUDB/Telemetry region
+				"X-Office-Platform": "Web",
+				"Content-Type": "application/json",
+			};
+			const correlationId = StringUtils.generateGuid();
```
ipdutta, line 57: Just for my information, what is this correlation id used for, since we are generating a new one everytime?

ArpanKundu97: As per https://www.owiki.ms/wiki/ICE_Services/Email_HRD, the correlation id is a unique GUID which will be used by the Email HRD team in their logs. If we ever need them to look-up their logs, we can give them a correlation id and a timestamp (preferably in UTC), and they can look up the logs.

It would have been best if we could have used the session id here. However, on using `Clipper.getUserSessionIdWhenDefined` here, clipper goes into an infinite loop. Hence, I have followed what ToDo does [here](https://dev.azure.com/outlookweb/Outlook%20Web/_git/to-do-web?path=/packages/todo-core/service/HRDService.ts&version=GBmaster&line=30&lineEnd=30&lineStartColumn=3&lineEndColumn=39&lineStyle=plain&_a=contents). We can re-visit this logic if there is ever a need for us to actually check the Email HRD logs.

ipdutta: In the code u have linked above, I see that the correlation id is getting logged hence in case of any failure or such we would be able to query it, but in our case that won't be possible since we are not logging it anywhere. If it's possible to log it, let's do that or else we can add a TODO and revisit it later as you mentioned.

ArpanKundu97: In the ToDo code, the correlation id is being logged using the functions `logger.info`, `logger.warn` and `logger.error`. These functions usually just log to the console in repositories owned by us, for example [here](https://office.visualstudio.com/OneNote/_git/sticky-notes-web-sdk?path=/packages/sticky-notes-sdk/src/lib/logging/logger.ts&version=GBmaster&line=79&lineEnd=87&lineStartColumn=3&lineEndColumn=4&lineStyle=plain&_a=contents). And looks like ToDo does the same, by looking at the code [here](https://dev.azure.com/outlookweb/Outlook%20Web/_git/to-do-web?path=/packages/todo-auth/src/core/logger.ts&version=GBmaster&line=26&lineEnd=37&lineStartColumn=3&lineEndColumn=4&lineStyle=plain&_a=contents).

But in any case, I have now added a comment to revisit the logic in future if need be.

ipdutta (approved):

PR #573, opened 2024-12-30
Fix for dark mode issues

ipdutta:
If we hide / delete the entire html element within the oneNoteWebClipper iframe, we can see a big white box which is what is hiding the contents of the actual webpage. The iframe has an inherited css property color-scheme as shown below:
image
However, clipper isn't being comfortably rendered in dark mode. On changing the value of the above property from dark to light, clipper works fine even when the search engine is in dark mode.

As per [color-scheme - CSS: Cascading Style Sheets | MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme), "The color-scheme CSS property allows an element to indicate which color schemes it can comfortably be rendered in".

Before: 
image
After:
image

Before:
image
After:
The white background does not appear in dark mode

### DIFF
--- a/src/scripts/extensions/styledFrameFactory.ts
+++ b/src/scripts/extensions/styledFrameFactory.ts
@@ -43,6 +43,7 @@ export class StyledFrameFactory {
 		iframe.style.position = "fixed";
 		iframe.style.transition = "initial";
 		iframe.style.zIndex = "2147483647";
+		(iframe.style as any).colorScheme = "light";
 	}
 
 	private static getGloballyStyledFrame(): HTMLIFrameElement {


src/scripts/extensions/styledFrameFactory.ts
```diff
@@ -43,6 +43,7 @@ export class StyledFrameFactory {
 		iframe.style.position = "fixed";
 		iframe.style.transition = "initial";
 		iframe.style.zIndex = "2147483647";
+		(iframe.style as any).colorScheme = "light";
```
SiddharthSinghMOD, line 46: ⁠how about declaring upfront that we only support light mode?

:root {
  color-scheme: light; // or only light
}

in our styles, hopefully it will not be overridden then?

ipdutta: The root component which has the colorScheme property, is an element of the original web page. Since our clipper html gets injected into the web page it inherits the CSS. To override this, we do not have the control to the root element. The clipper iframe is the topmost element of clipper which we have control over. Hence adding this CSS to the clipper iframe will be make the complete clipper UI stay in light mode, even when the actual webpage is in dark mode.

SiddharthSinghMOD (approved):

ArpanKundu97 (approved):

PR #583, opened 2025-03-03
Prevent XSS via usid

ArpanKundu97:
**Issue**
Reporter states that there is a Reflected XSS via the "usid" parameter. When the page is rendered, the ClipperSignOutUrl and ClipperFeedbackUrl variables are not being handled safely and therefore are escaped. This allows an attacker to jump straight to the hash location, which can be set as a javascript handler like "javascript:alert()". This leads to XSS.

**Fix**
Use a regular expression to ensure that the "usid" parameter conforms to expected patterns.

**Testing**
Hardcoded `usidQueryParamValue` in `getUserSessionIdQueryParamValue`.

`let usidQueryParamValue = "ee%22+location.replace(location.hash.substr(1))+%22&SilentAuthAttempted=1&fromAR=4#javascript:alert()";`

Verified that sign out does not go through and instead the following error is thrown in the console:

image

### DIFF
--- a/src/scripts/extensions/extensionWorkerBase.ts
+++ b/src/scripts/extensions/extensionWorkerBase.ts
@@ -340,7 +340,29 @@ export abstract class ExtensionWorkerBase<TTab, TTabIdentifier> {
 
 	protected getUserSessionIdQueryParamValue(): string {
 		let usidQueryParamValue = this.logger.getUserSessionId();
-		return usidQueryParamValue ? usidQueryParamValue : this.clientInfo.get().clipperId;
+		console.log("usidQueryParamValue: " + usidQueryParamValue);
+
+		// Validate the usid parameter (UUID-like format)
+		const uuidRegex = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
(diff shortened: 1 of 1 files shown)


ipdutta (approved):

src/scripts/extensions/extensionWorkerBase.ts
```diff
@@ -340,7 +340,26 @@ export abstract class ExtensionWorkerBase<TTab, TTabIdentifier> {
 
 	protected getUserSessionIdQueryParamValue(): string {
 		let usidQueryParamValue = this.logger.getUserSessionId();
-		return usidQueryParamValue ? usidQueryParamValue : this.clientInfo.get().clipperId;
+
+		// Validate the usid parameter (UUID-like format)
+		const uuidRegex = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+		const isValidUsid = uuidRegex.test(usidQueryParamValue);
+		if (!isValidUsid) {
+			throw new Error("Invalid usid parameter");
```
SiddharthSinghMOD, line 348: is there a way to query for these errors, do they get logged in some error telemetry?

SiddharthSinghMOD (approved):

PR #584, opened 2025-03-12
onenote.com subsites deprecation - Web Clipper Subsites

ArpanKundu97:
**Issue**
The following Web Clipper subsites are being deprecated:

1. https://www.onenote.com/whatsnext/webclipper
2. https://www.onenote.com/clipper/installed

**Fix**

https://www.onenote.com/whatsnext/webclipper - At this subsite, I only see change logs from very old versions from 2016, i.e. 3.1.0, 3.2.0 and 3.3.0. This indicates that the "What's New" tooltip might not even be shown to users currently. Hence, we have decided to disable the tooltip for now.

https://www.onenote.com/clipper/installed - Redirect to https://support.microsoft.com/en-us/office/getting-started-with-the-onenote-web-clipper-5696609d-c5ae-4591-b3af-1f897cb6eda6 on a successful clipper installation.

**Testing**
Verified that we are now successfully redirected to https://support.microsoft.com/en-us/office/getting-started-with-the-onenote-web-clipper-5696609d-c5ae-4591-b3af-1f897cb6eda6 on a successful clipper installation.

### DIFF
--- a/src/scripts/constants.ts
+++ b/src/scripts/constants.ts
@@ -377,10 +377,10 @@ export module Constants {
 		export var augmentationApiUrl = serviceDomain + "/onaugmentation/clipperextract/v1.0/";
 		export var changelogUrl = serviceDomain + "/whatsnext/webclipper";
 		export var clipperFeedbackUrl = serviceDomain + "/feedback";
-		export var clipperInstallPageUrl = serviceDomain + "/clipper/installed";
 		export var fullPageScreenshotUrl = serviceDomain + "/onaugmentation/clipperDomEnhancer/v1.0/";
 		export var localizedStringsUrlBase = serviceDomain + "/strings?ids=WebClipper.";
 
+		export var clipperInstallPageUrl = "https://support.microsoft.com/en-us/office/getting-started-with-the-onenote-web-clipper-5696609d-c5ae-4591-b3af-1f897cb6eda6";
 		export var msaDomain = "https://login.live.com";
 		export var orgIdDomain = "https://login.microsoftonline.com";
 		export var userDataBoundaryDomain = "https://odc.officeapps.live.com/odc/v2.1/federationprovider";
--- a/src/scripts/extensions/extensionBase.ts
+++ b/src/scripts/extensions/extensionBase.ts
@@ -153,11 +153,6 @@ export abstract class ExtensionBase<TWorker extends ExtensionWorkerBase<TTab, TT
 	protected getClipperInstalledPageUrl(clipperId: string, clipperType: ClientType, isInlineInstall: boolean): string {
 		let installUrl: string = Constants.Urls.clipperInstallPageUrl;
 
-		installUrl = UrlUtils.addUrlQueryValue(installUrl, Constants.Urls.QueryParams.clientType, ClientType[clipperType]);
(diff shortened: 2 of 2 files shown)


SiddharthSinghMOD (approved):

src/scripts/constants.ts
```diff
@@ -377,10 +377,10 @@ export module Constants {
 		export var augmentationApiUrl = serviceDomain + "/onaugmentation/clipperextract/v1.0/";
 		export var changelogUrl = serviceDomain + "/whatsnext/webclipper";
 		export var clipperFeedbackUrl = serviceDomain + "/feedback";
-		export var clipperInstallPageUrl = serviceDomain + "/clipper/installed";
 		export var fullPageScreenshotUrl = serviceDomain + "/onaugmentation/clipperDomEnhancer/v1.0/";
 		export var localizedStringsUrlBase = serviceDomain + "/strings?ids=WebClipper.";
```
sreealekhya-ms, line 381: Which all endpoints can we delete, are all of them in active usage?

src/scripts/constants.ts
```diff
@@ -377,10 +377,10 @@ export module Constants {
 		export var augmentationApiUrl = serviceDomain + "/onaugmentation/clipperextract/v1.0/";
 		export var changelogUrl = serviceDomain + "/whatsnext/webclipper";
 		export var clipperFeedbackUrl = serviceDomain + "/feedback";
```
sreealekhya-ms, line 379: This we should replace with other feedback mechanism, we should have something similar to: https://support.microsoft.com/en-us/office/send-us-your-feedback-about-onenote-for-mac-31683a4e-fc80-427d-a510-4e6eaa0e6e46

Check with Aditya once you can add me to that discussion

src/scripts/constants.ts
```diff
@@ -377,10 +377,10 @@ export module Constants {
 		export var augmentationApiUrl = serviceDomain + "/onaugmentation/clipperextract/v1.0/";
```
sreealekhya-ms, line 377: The onaugmentation ones seems like part of APIs, seems good for now, but if there is no actual usage let us delete these flows
https://dev.azure.com/office/Office/_git/Office?path=/oneservice/apis/Augmentation/onaugmentation/V1.0/WebApiConfig.cs&_a=contents&version=GBmain

ArpanKundu97: The `augmentationApiUrl` is used in the method `makeAugmentationRequest` in `augmentationHelper.ts`, which is used for rendering article previews. Hence, not deleting this.

src/scripts/constants.ts
ArpanKundu97: The `fullPageScreenshotUrl` is used in the method `getFullPageScreenshot` in `fullPageScreenshotHelper.ts`, which is used for rendering full page previews. Similarly, `augmentationApiUrl` is also in use as mentioned in my other reply.

The `localizedStringsUrlBase` is also in use in `localizationHelper.ts`. Hence, not deleting them.

src/scripts/constants.ts
ArpanKundu97: Sure, let me start the discussion on this. Will take this up as a follow-up.

PR #585, opened 2025-03-18
Update Feedback Mechanism

ArpanKundu97:
This PR is to update the feedback mechanism for OneNote Web Clipper so as to use https://feedbackportal.microsoft.com/feedback/post/c06dcc30-2e1c-ec11-b6e7-0022481f8472 instead of the https://www.onenote.com/feedback subsite.

**Testing**
Clicked on the `Feedback?` button and verified that the new feedback url pops up.

image

image

### DIFF
--- a/src/scripts/constants.ts
+++ b/src/scripts/constants.ts
@@ -376,11 +376,11 @@ export module Constants {
 
 		export var augmentationApiUrl = serviceDomain + "/onaugmentation/clipperextract/v1.0/";
 		export var changelogUrl = serviceDomain + "/whatsnext/webclipper";
-		export var clipperFeedbackUrl = serviceDomain + "/feedback";
 		export var fullPageScreenshotUrl = serviceDomain + "/onaugmentation/clipperDomEnhancer/v1.0/";
 		export var localizedStringsUrlBase = serviceDomain + "/strings?ids=WebClipper.";
 
 		export var clipperInstallPageUrl = "https://support.microsoft.com/en-us/office/getting-started-with-the-onenote-web-clipper-5696609d-c5ae-4591-b3af-1f897cb6eda6";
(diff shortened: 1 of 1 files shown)


src/scripts/constants.ts
```diff
@@ -376,11 +376,11 @@ export module Constants {
 
 		export var augmentationApiUrl = serviceDomain + "/onaugmentation/clipperextract/v1.0/";
 		export var changelogUrl = serviceDomain + "/whatsnext/webclipper";
-		export var clipperFeedbackUrl = serviceDomain + "/feedback";
 		export var fullPageScreenshotUrl = serviceDomain + "/onaugmentation/clipperDomEnhancer/v1.0/";
 		export var localizedStringsUrlBase = serviceDomain + "/strings?ids=WebClipper.";
 
 		export var clipperInstallPageUrl = "https://support.microsoft.com/en-us/office/getting-started-with-the-onenote-web-clipper-5696609d-c5ae-4591-b3af-1f897cb6eda6";
+		export var clipperFeedbackUrl = "https://feedbackportal.microsoft.com/feedback/post/c06dcc30-2e1c-ec11-b6e7-0022481f8472";
```
ipdutta, line 383: This looks like a different portal where the feedback will land - https://feedbackportal.microsoft.com. Are we not routing feedback to the specific browser web stores any more or this portal will do that behind the scenes?

ipdutta (approved):

src/scripts/constants.ts
ArpanKundu97: Users can still add their reviews to the browser web stores when prompted for a rating. When it comes to the Feedback feature though, users can now send their feedback to https://feedbackportal.microsoft.com/ instead of OCV.

PR #589, opened 2025-04-08
Remove webclipper/data usage

ArpanKundu97:
**Issue**
webclipper/data used in the function `sendMiscLogRequest` in WebClipper_Internal for telemetry purposes. We however want to deprecate this subsite.

**Fix**
`sendMiscLogRequest` is called only at 2 places.

1. sendNoOpTrackerRequest

We had disabled the clipper button on unclippable pages as part of https://github.com/OneNoteDev/WebClipper/pull/568. Hence, we shouldn't expect this function to be called anymore. Telemetry confirms the same.

image

Hence, I have commented out the call to `sendMiscLogRequest` in this function.

2. sendFailureLogRequest

I have replaced the call to `sendMiscLogRequest` with a call to `Clipper.logger.logFailure` in this function.

The `sendMiscLogRequest` function itself is being removed from WebClipper_Internal as part of https://github.com/OneNoteDev/WebClipper_Internal/pull/68.

**Testing**
Post merge, we will need to verify that the labels for which `sendFailureLogRequest` is called (mainly `UnhandledExceptionThrown`) continues getting logged.

### DIFF
--- a/src/scripts/logging/submodules/errorUtils.ts
+++ b/src/scripts/logging/submodules/errorUtils.ts
@@ -9,6 +9,7 @@ import {Localization} from "../../localization/localization";
 
 import {Failure, NoOp, unknownValue} from "../log";
 import { WebExtension } from "../../extensions/webExtensionBase/webExtension";
+import { Clipper } from "../../clipperUI/frontEndGlobals";
 
 export module ErrorUtils {
 	enum ErrorPropertyName {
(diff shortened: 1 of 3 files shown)


src/scripts/logging/submodules/errorUtils.ts
```diff
@@ -67,24 +68,17 @@ export module ErrorUtils {
 	 * Sends a request to the misc logging endpoint with relevant failure data as query parameters
```
ipdutta: Can you please update this comment as well? 
Also, if not already done it would be good to verify that the new logging won't break any of the alerting that we have since this is an error log.

ipdutta (dismissed):

src/scripts/logging/submodules/errorUtils.ts
ArpanKundu97: I have now updated the comment.

I had also checked the oneservice code and we are using the same Aria tenants as used by other OneNote Web Clipper events.

image

So even with `Clipper.logger.logFailure` we will continue logging the failure event to the same Aria tenant with the same label and hence this would not have any impact on alerting.

ipdutta (approved):

SiddharthSinghMOD (approved):

PR #651, opened 2026-03-26
Client-side migration: unified renderer window with self-contained sign-in

wikirby:
## Summary

A full client-side migration of the OneNote Web Clipper from the legacy injected Mithril sidebar to a standalone renderer window, with parity across all modes plus a Fluent 2 redesign.

### Architecture
- **Unified renderer window** replaces the injected Mithril sidebar with a standalone `renderer.html` (content iframe + sidebar) opened by the service worker
- **Self-contained sign-in** — renderer shows MSA/OrgId sign-in panel when not authenticated; worker opens OAuth popup; no more `clipperInject.ts` injection in the primary flow
- **Client-side capture pipeline** — standalone `contentCaptureInject.ts` performs full DOM cleaning (master-compatible DomUtils baseline + enhancements: hidden element preservation, position neutralization, shadow DOM flattening, lazy image resolution)
- **Direct OneNote save** — worker builds multipart form and POSTs directly to OneNote API (no server proxy)
- **Chunked port streaming** — images sent via per-port `saveImage` chunks instead of session storage (multi-window safe, no 10MB quota bottleneck)

### Modes
- **Full Page** — per-viewport `captureVisibleTab` with sidebar cropping, JPEG 95% stitching in renderer
- **Article** — Readability via dynamic import, ONML cleanup for OneNote page styling, preview header with highlighter / serif-sans toggle / font size controls
- **Bookmark** — og:description/og:image extraction with full legacy fallback chain (keywords, article:tag, page text). Thumbnail converted to base64 data URL via canvas (matches legacy `DomUtils.getImageDataUrl`) so OneNote can render it (it cannot fetch external URLs)
- **Region** — standalone `regionOverlay.ts` with Shadow DOM instruction bar, crosshair overlay, multi-region thumbnails. Keyboard support: arrow keys with velocity acceleration, two-phase Enter, ARIA live announcements, seamless keyboard→mouse handoff
- **PDF** — `pdf.combined.js` integration; lazy-loaded preview pages, page range selection (with validation), distributed-pages mode (one OneNote page per PDF page), attachment support up to 24.9MB, per-page progress reporting

### Fluent 2 Design
Per designer Figma spec (`wK4ryPaULoiSDMt2xVc1fH`):
- White Fluent theme replaces dark purple sidebar
- OneNote brand ramp (`Brand/OneNote/50/60/80` from Fluent 2 library) for primary button states (rest/hover/pressed)
[...]

PR #658, opened 2026-05-14
Article-mode oEmbed extraction for video pages + V1 payload parity

wikirby:
## Summary

V1's article-mode flow on YouTube/Vimeo produced a saved OneNote page with an embedded video iframe, citation, title/author caption, and `AutoPageTags*` meta tags that OneNote's renderer uses to recognize the result as an article-style clip with playable embeds. V2 shipped without that machinery — clicking Article on a YouTube page just ran Readability, which strips iframes and produces a text-only result with no player and no description. Users reported the regression.

This PR restores V1's user-visible behavior using the [oEmbed standard](https://oembed.com/) instead of V1's per-provider URL parsing. Both YouTube and Vimeo publish CORS-enabled oEmbed endpoints that the chrome-extension origin can fetch directly under existing `<all_urls>` host_permissions.

## Provider scope

Narrowed to V1 parity: **YouTube + Vimeo** only.

V1 also handled Khan Academy via regex scrape of lesson-page HTML for embedded YouTube IDs — markup likely no longer matches modern Khan Academy pages; skipped per maintainer direction. Additional providers may be added as a focused capability-expansion PR in the future.

## Changes

| Area | Change |
|---|---|
| `src/scripts/contentCapture/oembedExtractor.ts` (new) | 3-entry provider table + hostname-pattern matching + fetch/parse + `sanitizeProviderHtml` helper. Returns raw `OEmbedData` (renderer composes preview & save separately). |
| `renderer.ts` `extractArticle` | Try oEmbed first; on null fall through to existing Readability path (no behavior change for non-matching domains). |
| `renderer.ts` `composeOEmbedForPreview` | Thumbnail at fixed 600x338 (16:9 — matches saved iframe size), title `<h3>`, "author · provider" attribution, og:description, CSS-only `▶` overlay on video-type thumbnails. No nested iframe in preview because preview-frame is sandboxed (`allow-same-origin`) and the player can't execute JS inside it. |
| `renderer.ts` `composeOEmbedForSave` | Sanitized provider iframe with `data-original-src=<pageUrl>` injected and dimensions normalized to 600x338 — the marker OneNote's renderer uses to recognize the embed (matches V1's `YoutubeVideoExtractor.createEmbeddedVideoFromId`). |
| `renderer.ts` save dispatch | Threads new `pageMetadata` map through the save port message. Article paths (oEmbed + Readability) populate `AutoPageTagsCodes=Article`, `AutoPageTags=Article`, plus title/author/siteName + description. |
| `webExtensionWorker.ts` `buildPage` | HTML output realigned to V1 `OneNoteApi.OneNotePage.getEntireOnml`: no `<!DOCTYPE>`, `<html xmlns="http://www.w3.org/1999/xhtml" lang=<locale>>` (no quotes around lang — matches V1 literally), iterates `pageMetadata` to emit one `<meta>` per entry. Same shape applied to the parallel `distHtml` builder for distributed-PDF saves. |
| `renderer.ts` `imageToDataUrl` | Drive-by bookmark fix: initial PNG encode + iterative JPEG-quality step-down when encoded data URL exceeds OneNote's per-MIME-part limit. Matches V1's deleted `DomUtils.adjustImageQualityIfNecessary`. Surfaced because bookmark mode on YouTube was hitting `400 Maximum request size exceeded` (1280x720 og:image PNG-encoded ~2.5MB). |

## Follow-up fixes (separate commits)

- **Article-mode highlighter** (`9bc0549`): Pre-existing regression from PR #651's sandbox hardening. Library now loads as a parent-window script (same pattern as pdf.js) and operates on the iframe DOM via `allow-same-origin` rather than executing inside it. Small `textHighlighter.js` patch routes event bindings through `el.ownerDocument.defaultView`. Sandbox stays tight.
- **Sidebar group spacing** (`ca6ac23` + `a292926`): Per designer, more breathing room around the Fluent dividers — `.sidebar-group` padding 8px → 12px.
- **Segoe UI @font-face restored for Mac/Linux** (`fe40b3d`): Re-adds the V1 `@font-face` declaration so non-Windows users get Segoe UI from OneNote's CDN instead of falling through to `-apple-system` (San Francisco). Local() cascade keeps Windows on the installed font, no network request. No CSP change — font loading was already unrestricted.
- **Selection mode + ContextImage + cross-mode highlight preservation** (`be5b15c`): Restores V1's right-click context-menu features that were either no-ops or broken in V2.
  - **Listener bug fix**: `contextMenus.onClicked.addListener` was registered inside the menu-creation `for` loop, so a single right-click fired the handler N times — N parallel `captureVisibleTab` requests blew the per-second quota and the popup blinked closed. Hoisted out of the loop.
  - **"Clip Image to OneNote"** (ContextImage) — previously dispatched but ignored by V2's renderer. Worker now forwards `invokeDataForMode` (image srcUrl) to the renderer, which fetches via the existing `imageToDataUrl` and pre-seeds Region mode with the single image as a thumbnail (matches V1's `regionResult.data = [srcUrl]` + default-mode-Region pattern).
  - **"Clip Selection to OneNote"** — new 6th conditional mode button revealed only when invoked via right-click. Selection HTML capture in `contentCaptureInject.ts` reuses the existing full-DOM cleanup pipeline by substituting `body` with the cloned range, materializing `img.src`/`a.href` from properties so absolute URLs survive into `body.innerHTML`. Result piped through `cleanArticleHtml` for ONML-equivalent strip — `iframe` added to the strip list (safe because the oEmbed path bypasses `cleanArticleHtml` by construction, so video iframes for known providers still flow through). Save reuses article's branch with `AutoPageTags=Article` (V1 parity). Button uses Fluent UI System Icons; label/tooltip wired via existing V1 i18n keys (no new strings).
  - **Cross-mode highlight preservation** — pre-existing latent bug: `articleWorkingHtml` saved on switch-out of article, but `switchToArticle` / `switchToSelection` themselves didn't capture the *outgoing* mode's state, so article ↔ selection ↔ article would have lost article's highlights. Generalized `saveArticleWorkingState` → `saveWorkingState` dispatched by `currentMode`, added the save call at every transition entry point, added `selectionWorkingHtml` slot.
  - **Highlighter line-jump in `<pre>` blocks** — `.highlight-anchor` was `display: inline-block`, making the wrapped span an atomic layout unit and defeating `word-break: break-word`. Long highlighted URLs in code blocks jumped to their own line. Changed to `display: inline`.
- **Synchronous context-menu listener registration** (`70b0fb1`): MV3 SW unloads aggressively (~30s). Listener now registers at SW startup instead of after `clipperIdProcessed`, so right-clicks during a wake-up aren't dropped. Menu items themselves still register async because their titles need locStrings.
- **Skip eager full-page capture on context-menu invocations** (`4db0d29`): Selection and Image context-menu invokes don't need a full-page screenshot; deferred until/unless the user switches into Full Page mode.
- **Open in Article mode on known oEmbed pages from toolbar invoke** (`59a4f25`): YouTube/Vimeo toolbar invocations now auto-engage Article mode (the user's likely target) instead of dropping into Full Page first.
- **Storage permission cleanup** (`d343f64`): Drops the `"storage"` permission from manifests — V3 renderer uses `localStorage` (shared extension origin), no `chrome.storage` usage remains.
- **Three deferred-capture polish fixes** (`ea8f1d9`): Three small issues surfaced by the dropdown + deferred-capture work — see commit message for specifics.
- **Mode-selector dropdown attempt + revert** (`6af8a0d` + `2c9d42b`): A Fluent 2 dropdown collapsing the 5–6 stacked mode buttons was prototyped, but the designer asked for all buttons up front. Reverted; window-height cap bumped from 900 → 980 to comfortably fit all buttons + PDF mode controls without scroll on standard displays.
- **PDF-attach visual indicator** (`7b84921`): Reviewer flagged that V1 showed a PDF icon at the top of the preview when "Attach PDF file" was checked. Restored via V1's `PdfPreviewAttachment` (84x96 icon + filename, top-left of preview), re-mounts on every previewContainer rebuild.
- **Drop unused "cookies" permission from Edge manifest** (`6936290`): Chrome manifest never declared it; Edge had a residual entry. V3 never reads `document.cookie` or calls `chrome.cookies`, so removed.
- **Refresh user state on renderer open** (`6566499`): V1's `getInitialUser` ran `auth.updateUserInfoData(InitialRetrieval)` on every clipper open — cache-aware via TTL, so no network call when the cached token is fresh. V2 was missing this hook, so stale tokens surfaced as a notebooks-401 wave on open. Renderer now sends `refreshUser` to the worker on boot; worker fires `updateUserInfoData` and pings back `userRefreshed`, which triggers `fetchFreshNotebooks`. No new auth UI — purely a token-freshness path.
- **Selection-mode telemetry parity + comment cleanup** (`2529b97`, `2dd9340`): Selection-mode button included in `modeClickIds` map for telemetry parity with other mode buttons; verbose multi-line "why" comments across PR-touched files trimmed to one-liners or removed where the code is self-explanatory (production bundle is minified, so this is purely a review-surface change).

## Test plan

- [x] `npm run build:prod` clean (zero TS errors, tslint passes)
- [x] YouTube watch page → article mode → preview shows thumbnail + ▶ overlay + title/author/description; save produces OneNote page with playable embedded video
- [x] Vimeo video page → same flow
- [x] Non-matching domain (regular article) → falls through to Readability with no regression
- [x] Bookmark mode on YouTube → saves successfully (was 400 before the `imageToDataUrl` JPEG fallback)
- [x] Article-mode highlighter: select text → yellow span with red-circle delete button; click delete removes
- [x] Segoe UI loads from network on machines without it locally
- [x] Right-click "Clip Image to OneNote" → opens in Region mode with that image pre-seeded as a thumbnail
- [x] Right-click "Clip Selection to OneNote" → opens with Selection mode button revealed and auto-engaged; selected text preserved with absolute image/link URLs; iframes stripped
- [x] Highlights preserved across article ↔ selection ↔ article (and via other modes)
- [x] Long URL highlighted in a `<pre>` block wraps mid-token instead of jumping to a new line
- [x] PDF mode: "Attach PDF file" checkbox shows V1-style icon+filename indicator at top-left of preview
- [x] YouTube/Vimeo toolbar invoke → opens directly in Article mode (skip Full Page)
- [x] Sign-in → token refresh on renderer open → notebooks fetch succeeds without 401 wave
- [x] PDF mode, full-page mode, region mode otherwise unaffected

## Out of scope (deferred)

- Image-size fallback on PDF mode / full-page mode / region mode — those paths already use JPEG and aren't reported as failing
- Broader provider support — separate, focused PR if appetite arises
- Pixel-perfect Mac/Windows parity on Semibold rendering (browser synthesizes from Regular today; add `@font-face` for `Segoe UI Semibold` if designer flags it)
- V1's video-domain iframe special-handling in selection mode (V1 kept YouTube/Vimeo iframes in selections via `VideoExtractorFactory`; V2 strips all iframes from selection — users who want a video clip use article mode)
- Multi-window sign-out propagation (vanishingly rare; self-bounded by 401 on next action)

## Demo
image
image

🤖 Generated with [Claude Code](https://claude.com/claude-code)


aanchalbhansali:
@wikirby There are too many long-liner comments in all the changed files. Please clean them up / shorten.

src/scripts/renderer.ts
aanchalbhansali: Please add constant for "selection" here for telemetry parity

wikirby: resolved in latest commit

aanchalbhansali (approved):
Approved

PR #660, opened 2026-05-22
Dependency upgrade and V1 dead-code cleanup

wikirby:
## Summary
- Removes V1-era source files (Mithril sidebar wiring, page-nav tooltip system, video extractor pipeline, polyfills, etc.) and their npm packages
- Upgrades the build toolchain: gulp 3 → 5, TypeScript 3.5 → 5.7, pdfjs-dist 1.7 → 5.7, jwt-decode 2 → 4, plus minor majors across the gulp ecosystem
- Pins a set of transitive dev-deps via npm overrides
- Bumps @types/chrome 0.0.257 → 0.1.42 (now possible with TS 5+) and renames the chrome event-info types accordingly
- Vendors the section-picker PNGs into `src/images/` (previously copied from `onenotepicker/target/images/`, which is now removed)
- One source-file case fix: `dataBoundary.ts` → `DataBoundary.ts` to match the internal repo's import
- **Round 2 (below):** drops `browserify` / `gulp-typescript` / `gulp-cssnano` / `onenoteapi` / `tslint`, replaces with `esbuild` / direct `tsc` / `gulp-postcss + cssnano 8` / local types shim / ESLint flat config; bumps JS target to ES2022 with a first modernization sweep
- **Round 4 (below):** relaxes the mode-button lock that V3 imposed during initial FullPage capture so users can pivot to Article / Bookmark / Region without waiting for the screenshot loop to finish (V1 parity)

## Round 2: Build infra modernization

- Replace `browserify` with `esbuild` for bundling. Each entry point becomes an `esbuild.build({ format: "iife", ... })` call; the LogManager UMD `standalone` wrap is handled by esbuild's `globalName` option. Bundle sizes net -71KB (-7.2%) from better tree-shaking. Drops the unused `crypto-browserify` / `elliptic` polyfill chain that browserify auto-pulled.
- Drop `gulp-typescript` (6+ years stale; no v6 stable, latest is `6.0.0-alpha.1` from Nov 2019). Compile step becomes a `child_process.spawn` of local `typescript/bin/tsc -p tsconfig.json`. Output byte-identical to before.
- Replace `gulp-cssnano` with `gulp-postcss` + `cssnano 8` + `postcss 8` (the modern decomposition recommended by cssnano's own README since 2018). Build-time only; renderer.css minifies to the same ~17% reduction.
- Drop `onenoteapi` runtime dep. V3 calls the OneNote API directly via `fetch` already; the only consumed surface was the `RequestErrorType` enum + a few type names. Replaced with a 27-line local `src/scripts/oneNoteApi.ts` shim imported explicitly by the 17 consumers (also drops the unused `IOneNoteApi` interface from `responsePackage.ts` and the gulpfile concat that injected `oneNoteApi.min.js` into the SW bundle).
- Migrate `tslint` → `ESLint 10 + typescript-eslint 8` (flat config in `eslint.config.mjs`). tslint went end-of-life in 2019; the 23 rules from `src/tslint.json` are translated to ESLint equivalents. Code touch-ups: `var` → `let` across 18 files (262 occurrences in the three namespace files alone), four surgical `// eslint-disable-next-line no-console` at legitimate call sites + one file-level disable on `webConsole.ts` (its entire purpose is console wrapping), indent + trailing-space `eslint --fix`. The gulp `tslint` task is replaced by a `lint` task (spawn-based eslint) and kept as an alias so external callers / pipeline refs keep working.
- Fix prod build under gulp 5: `mergeSettings` passes `allowEmpty: true` to `gulp.src`. Gulp 5 errors hard on a singular glob that doesn't resolve; gulp 4 silently emitted no files for missing overrides. The override files (`production.json` / `dogfood.json`) live in `WebClipper_Internal`, not the public repo, so they were tripping `--production` builds since the gulp 5 upgrade.

## Round 3: ES2022 + first modernization sweep

- Bump JS target ES2017 → ES2022 across tsconfig (`target` + `lib`), gulpfile (esbuild `target`), eslint config (`ecmaVersion`), and both manifests (`minimum_chrome_version: "94"`). Chromium 94 (Sep 2021) is the floor; covers private class fields, top-level await, `Object.hasOwn`, `.at()` on arrays/strings, and logical assignment operators. MV3's hard minimum is Chromium 88 — real-world auto-updated installs are far ahead.
- Enable typed linting (`parserOptions.project` pointing at tsconfig.json) and the `@typescript-eslint/prefer-optional-chain` rule. First sweep covers 14 source files: `x && x.y` (and `x && x.y ? x.y : z` ternary forms) → `x?.y` / `x?.y ?? z`. Mostly `eslint --fix`; ~25 ternary cases the rule won't autofix were done by hand. All semantically equivalent for our LHS shapes (objects/functions, never `0` / `false` / `""` sensitive).
- One `Object.prototype.hasOwnProperty.call(o, k)` → `Object.hasOwn(o, k)` (new in ES2022).

Followup typed-linting rules (`prefer-nullish-coalescing`, `prefer-includes`, `prefer-string-starts-ends-with`, `no-floating-promises`) and structural cleanup of the larger V3 worker/renderer files are deferred to a focused successor PR — scope discipline for review.

Dropped from devDependencies: `browserify`, `gulp-typescript`, `gulp-cssnano`, `gulp-plumber`, `gulp-tslint`, `tslint`, `vinyl-source-stream`, `merge-stream`, `onenoteapi`.
Added: `esbuild`, `eslint`, `typescript-eslint`, `cssnano`, `gulp-postcss`, `postcss`.

Renamed: `src/scripts/logging/logManager.d.ts` → `logManagerGlobals.d.ts`. The legacy `gulp.src("**/*.+(ts|tsx)")` glob happened to include `.d.ts` files (suffix is `.ts`), masking the fact that tsc's default project discovery treats a co-named `.d.ts` as the auto-generated declaration for its `.ts` sibling and drops it from inputs. Renaming makes the ambient `LogManager` global visible again.

## Round 4: Mode switching during initial capture (V1 parity)

V3 disabled all non-FullPage mode buttons until the screenshot loop completed. V1 allowed switching during capture; this restores that UX.

- **Drop the three mode-button disable loops** (initial lock, `switchToFullPage` re-entry, sign-out reset). Save + signout still lock during capture — those have legitimate race-condition concerns; mode switching does not, because the capture is worker-driven and async.
- **New `abortCurrentCapture()` helper** in renderer.ts centralizes the cancel teardown: sends `cancelCapture` to the worker, flips the local `captureInProgress` flag, clears the progress UI, resets `stitchYOffset`, and calls `showPreviewFrame()` to restore the rounded frame overlay. Called from the click handler when the user leaves FullPage mid-capture (replacing a previous early-return that blocked the switch).
- **Worker-side `captureCancelled` flag** scoped to `setupPort`, set on `cancelCapture` and reset on `dimensions`. Four async-resume points (the `scrollResult` setTimeout, the `windows.update` callback, the `captureVisibleTab` callback, the `drawComplete` handler entry) short-circuit on the flag, so the worker stops the scroll/captureVisibleTab loop at the source rather than the renderer having to filter stale messages.
- **`switchToFullPage`'s re-entry path defers `kickoffFullPageCapture` through two nested `requestAnimationFrame` calls**, with an explicit `iframe.contentWindow.scrollTo(0, 0)` after a forced reflow. The double rAF is load-bearing: when the iframe transitions `display:none` → `display:block`, Chrome needs at least one paint cycle to commit the display change and another for the scroll position. Without the defer, the worker's `captureVisibleTab` (which fires ~500 ms later in its own `setTimeout`) grabs the iframe at its prior scroll position and the restarted capture appears to "continue where it left off". One rAF tested intermittently; two rAFs is consistently reliable.
- **`finalize`'s UI updates guard on `currentMode === "fullpage"`** so a capture that completes while the user has already pivoted to another mode doesn't force-enable Save or steal focus back to the FullPage button.

Net: ~50 lines across renderer.ts and webExtensionWorker.ts.

## Notable behavioral changes (none expected at runtime)
- pdfjs-dist v5 is ESM-only, so a small `src/pdfLoader.mjs` shim now imports pdf.mjs and exposes it on `window.pdfjsLib`. renderer.ts waits on a `pdfjs-ready` event before calling getDocument. PDF mode end-to-end smoke test is the main regression risk.
- gulp 5's streamx engine doesn't compose `merge-stream` pipelines well; the gulpfile now uses `Promise.all` + a small streamToPromise helper for parallel work and `gulp.series` for sequencing (replaces `run-sequence`).
- jwt-decode v4 changed from default export to named export — the one call site in `cachedHttp.ts` was updated.
- ESLint `no-console` is now actually enforced. The four pre-existing intentional call sites (offscreen warn paths, public-stub logManager, webConsole wrapper) have surgical disables.
- Mode buttons are clickable during initial FullPage capture. Clicking away cancels the in-flight capture cleanly; returning to FullPage restarts the capture from scratch (V1 parity).

## Test plan
- [ ] Sign in (MSA + OrgId)
- [ ] Full Page clip
- [ ] Article mode clip (incl. highlighter + font toggles)
- [ ] Region mode clip
- [ ] Bookmark mode clip
- [ ] PDF mode — `https://` PDF and `file:///` PDF
- [ ] Sign out + sign back in
- [ ] **Mode switching during initial capture** — launch clipper on a multi-tile page, click Region (or Article, Bookmark) while capture is in progress, verify the switch works
- [ ] **Mode-switch return to FullPage** — from the above, return to FullPage; verify capture restarts from scratch and stitches the full page (no stale "continues where left off" behavior)
- [ ] **Esc out of Region without clip** — click Region mid-capture, press Esc, verify snap-back to FullPage restarts capture cleanly
- [ ] Verify Aria telemetry still flushing (Network tab for `browser.events.data.microsoft.com`)
- [ ] `npm run build` clean
- [ ] `npm run build:prod` clean (with `WebClipper_Internal` sibling clone)
- [ ] CI build green


aanchalbhansali (approved):
Approved

PR #661, opened 2026-06-05
Restore collapsed-by-default location dropdown

wikirby:
## Problem

The location (notebook/section) dropdown rendered every notebook and section group fully expanded. Customers with many notebooks/section groups had to scroll through the entire flattened tree to find their target section.

In the original picker, the dropdown defaulted to **collapsed** and expanded only the notebook and section groups leading to the currently selected section. The V2/V3 unified-window rewrite of the picker lost that behavior.

## Fix

Restore the collapse-by-default behavior in [renderer.ts](src/scripts/renderer.ts):

- `flattenSections` / `flattenSectionGroups` now report whether their subtree contains the selected section, so each notebook/section-group heading collapses unless it sits on the path to that section.
- New `refreshSectionVisibility()` recomputes row visibility from collapsed state in a single pass that tracks the shallowest collapsed depth, so nested collapse/expand is honored correctly. Section rows now carry `data-depth`.
- When there is no stored selection (or it no longer exists), default to the first available section and reveal its path, so the dropdown always opens to a visible, expanded selection rather than an all-collapsed list.

## Notes

- On open, the selected section is focused and scrolled into view as before; only its ancestry is expanded. Other notebooks/groups can be expanded by clicking their chevron.
- Single-file change. Build and lint pass; verified the change ships in the Edge target.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/scripts/renderer.ts
+++ b/src/scripts/renderer.ts
@@ -523,13 +523,15 @@ function populateSectionDropdown() {
 				curSectionId = cur.section ? cur.section.id : "";
 			} catch (e) { /* ignore */ }
 		}
+		// Fall back to the first section so the dropdown opens to a visible, expanded path.
+		let hasStoredSelection = !!curSectionId && sectionExists(notebooks, curSectionId);
+		if (!hasStoredSelection) {
+			curSectionId = findFirstSectionId(notebooks);
+		}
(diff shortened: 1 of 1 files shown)


aanchalbhansali (approved):
Approved

PR #667, opened 2026-07-06
Enable Sign out during initial FullPage capture

wikirby:
## Summary

On opening the clipper, **Sign out** was disabled while the initial full-page capture ran, and only re-enabled once that capture completed. Since mode buttons stay interactive during capture (added in "Allow mode switching during initial FullPage capture"), a user could switch to Article/Selection/Bookmark/Region before the capture finished — and in those modes Sign out was never re-enabled, leaving it stuck disabled.

## Root cause

`disableSignout()` fired on dialog open and in `switchToFullPage()`. Sign out was only re-enabled by capture completion (`finalize`), PDF mode, or the no-capture invoke path — none of which run when the user pivots to another mode mid-capture.

## Fix

- Remove the capture-time `disableSignout()` calls. Sign out now locks **only during the save round-trip**, via `lockSidebar()` / `unlockSidebar()`.
- Remove the now-redundant standalone `enableSignout()` calls (PDF mode, no-capture ready state, finalize, sign-out reset), so Sign out state is driven entirely by lock/unlock.
- Abort any in-flight capture when the user signs out, so the worker's capture loop doesn't outlive the session.

Net invariant: Sign out state == save-lock state, always.

## Testing

- [x] `npm run build` clean (compile + lint)
- [x] Open clipper, switch modes mid-capture -> Sign out stays available
- [x] Save round-trip -> Sign out disabled during save, re-enabled after
- [x] Sign out mid-capture -> capture aborts cleanly, returns to sign-in panel

### DIFF
--- a/src/scripts/renderer.ts
+++ b/src/scripts/renderer.ts
@@ -790,6 +790,8 @@ let feedbackLabel = document.getElementById("feedback-label");
 if (feedbackLabel) { feedbackLabel.textContent = strings.feedback; }
 signoutLink.addEventListener("click", (e) => {
 	e.preventDefault();
+	// Stop any in-flight full-page capture so the worker's loop doesn't outlive the session.
+	abortCurrentCapture();
 	logFunnel(Funnel.Label.SignOut);
 	logSessionEnd(Session.EndTrigger.SignOut);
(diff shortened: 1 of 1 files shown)


aanchalbhansali (approved):
Approved

PR #671, opened 2026-09-03
feat(firefox): add MV3 compatibility

s-n-d-p:
## Summary
- add a Firefox MV3 build and XPI package using `background.scripts`
- preserve the historical Firefox add-on identity
- fall back when Chrome offscreen APIs are unavailable

## Validation
- `npm run build:prod`
- Firefox package lint: 0 errors
- loaded the generated XPI in Firefox

### DIFF
--- a/src/scripts/communicator/offscreenCommunicator.ts
+++ b/src/scripts/communicator/offscreenCommunicator.ts
@@ -35,6 +35,10 @@ export async function sendToOffscreenDocument(type: string, data: any): Promise<
 	let chromeRuntime = chrome.runtime as any;
 	let chromeOffscreen = (chrome as any).offscreen;
 
+	if (typeof chromeRuntime.getContexts !== "function" || !chromeOffscreen) {
+		return handleWithoutOffscreenDocument(type, data);
+	}
+
 	const existingContexts = await chromeRuntime.getContexts({
 		contextTypes: [chromeRuntime.ContextType.OFFSCREEN_DOCUMENT],
 		documentUrls: [offscreenUrl]
@@ -63,4 +67,27 @@ export async function sendToOffscreenDocument(type: string, data: any): Promise<
 			});
 		});
 	});
+}
+
+function handleWithoutOffscreenDocument(type: string, data: any): string {
+	switch (type) {
+		case OffscreenMessageTypes.getFromLocalStorage:
+			return localStorage.getItem(data.key);
+		case OffscreenMessageTypes.setToLocalStorage:
+			localStorage.setItem(data.key, data.value);
+			return "SUCCESS";
+		case OffscreenMessageTypes.removeFromLocalStorage:
+			localStorage.removeItem(data.key);
+			return "SUCCESS";
+		case OffscreenMessageTypes.getHostname: {
+			let url = new URL(data.url, offscreenUrl);
+			return url.protocol + "//" + url.host + "/";
+		}
+		case OffscreenMessageTypes.getPathname:
+			return new URL(data.url, offscreenUrl).pathname;
+		default:
+			// eslint-disable-next-line no-console -- match the offscreen handler's failure mode
+			console.warn(`Unexpected message type received: '${type}'.`);
+			return;
+	}
 }
--- a/src/scripts/extensions/firefox/firefoxExtension.ts
+++ b/src/scripts/extensions/firefox/firefoxExtension.ts
@@ -1,0 +1,6 @@
+import {ClientType} from "../../clientType";
+import {WebExtension} from "../webExtensionBase/webExtension";
+
+// Firefox exposes the callback-based chrome namespace used by the shared implementation.
+WebExtension.browser = chrome;
+let clipperBackground = new WebExtension(ClientType.FirefoxExtension);
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -26,6 +26,7 @@ var PATHS = {
     TARGET: {
         ROOT: "target/",
         CHROME: "target/chrome/",
+        FIREFOX: "target/firefox/",
         EDGE_ROOT: "target/edge/OneNoteWebClipper/edgeextension/",
         EDGE_EXTENSION: "target/edge/OneNoteWebClipper/edgeextension/manifest/extension/"
     },
@@ -207,6 +208,10 @@ gulp.task("bundleChrome", function () {
     return bundleEntry(PATHS.BUILDROOT + "scripts/extensions/chrome/", "chromeExtension.js");
 });
 
+gulp.task("bundleFirefox", function () {
+    return bundleEntry(PATHS.BUILDROOT + "scripts/extensions/firefox/", "firefoxExtension.js");
+});
+
 gulp.task("bundleEdge", function () {
     return bundleEntry(PATHS.BUILDROOT + "scripts/extensions/edge/", "edgeExtension.js");
 });
@@ -219,6 +224,7 @@ gulp.task("bundle", gulp.series(
     "bundleRenderer",
     "bundleLogManager",
     "bundleChrome",
+    "bundleFirefox",
     "bundleEdge"
 ));
 
@@ -235,6 +241,7 @@ function lowerCasePathName() {
 
 var targetDirHasExportedCommonJs = {};
 targetDirHasExportedCommonJs[PATHS.TARGET.CHROME] = false;
+targetDirHasExportedCommonJs[PATHS.TARGET.FIREFOX] = false;
 targetDirHasExportedCommonJs[PATHS.TARGET.EDGE_EXTENSION] = false;
 function exportCommonJS(targetDir) {
     if (targetDirHasExportedCommonJs[targetDir]) {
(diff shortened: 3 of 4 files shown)


src/scripts/communicator/offscreenCommunicator.ts
```diff
@@ -64,3 +68,22 @@ export async function sendToOffscreenDocument(type: string, data: any): Promise<
...
+		case OffscreenMessageTypes.getFromLocalStorage:
+			return localStorage.getItem(data.key);
+		case OffscreenMessageTypes.setToLocalStorage:
+			localStorage.setItem(data.key, data.value);
+			return "SUCCESS";
+		case OffscreenMessageTypes.removeFromLocalStorage:
+			localStorage.removeItem(data.key);
+			return "SUCCESS";
+		case OffscreenMessageTypes.getHostname:
+			return new URL(data.url, offscreenUrl).origin + "/";
```
ipdutta: **[Resolved — non-blocking, no action needed]** ~~`URL.origin` and the anchor-based implementation disagree on non-special schemes.~~

I raised this because the offscreen implementation builds the hostname as `l.protocol + "//" + l.host + "/"` from an anchor element, whereas this fallback uses `new URL(...).origin + "/"`. Those agree for `http:`/`https:` (port included), but `URL.origin` is the literal string `"null"` for opaque-origin schemes, so a `file://` URL would yield `"null/"` here where Chrome yields `"file:///"`.

**I've since checked, and it isn't reachable.** Both manifests gate host access on `<all_urls>`, which does not cover `file://` — Firefox requires an explicit `file:///*` match pattern, and Chrome puts it behind the user's "Allow access to file URLs" toggle. Since `handleWithoutOffscreenDocument` only ever executes on Firefox, a `file://` URL can't reach `getHostname` on that path. `data:`/`blob:` top-level pages aren't clippable either.

Leaving this here as a note rather than a request: if `file:///*` is ever added to the match patterns, this line needs to mirror the anchor logic (`u.protocol + "//" + u.host + "/"`). Nothing to do now — apologies for the detour.

src/scripts/communicator/offscreenCommunicator.ts
```diff
@@ -64,3 +68,22 @@ export async function sendToOffscreenDocument(type: string, data: any): Promise<
...
+		case OffscreenMessageTypes.getPathname:
+			return new URL(data.url, offscreenUrl).pathname;
+		default:
+			throw new Error(`Unexpected offscreen message type: '${type}'.`);
```
ipdutta: **[Low]** Divergent failure mode for unknown message types.

This fallback throws, which rejects the promise returned by `sendToOffscreenDocument`. The offscreen path `console.warn`s and returns `false`, so the caller resolves `undefined` instead. Identical bad input therefore breaks a caller on Firefox and silently no-ops on Chrome.

The throw is arguably the better behaviour, but having it apply on only one browser is the worst of both. I'd match the existing `console.warn` + return here, and if the added strictness is wanted, change both together.

src/scripts/extensions/firefox/firefoxExtension.ts
```diff
@@ -0,0 +1,6 @@
+import {ClientType} from "../../clientType";
+import {WebExtension} from "../webExtensionBase/webExtension";
+
+// Firefox exposes the callback-based chrome namespace used by the shared implementation.
+WebExtension.browser = chrome;
```
ipdutta, line 5: **[Low]** Asymmetry with the chrome/edge entry points.

`chromeExtension.ts` and `edgeExtension.ts` both set `WebExtension.offscreenUrl` alongside `WebExtension.browser`; this file sets only `browser`.

I traced it before assuming a bug: `WebExtension.offscreenUrl` is declared in `webExtension.ts` and assigned in those two files, but never read anywhere — `offscreenCommunicator` uses its own module-local `offscreenUrl`. So omitting it is harmless, and arguably more correct. Since it reads as an oversight, either a one-line comment here or deleting the now-dead static from `webExtension.ts` would settle it.

One broader note on the new guard in `sendToOffscreenDocument`: `typeof chromeRuntime.getContexts !== "function"` is true in *any* context that doesn't expose `getContexts`, not just Firefox — including Chrome content scripts. I checked, and `contentCaptureInject.ts`, `regionOverlay.ts` and `appendIsInstalledMarker.ts` are all standalone IIFEs with no imports, so nothing reaches this today. But the fallback would use the *page's* `localStorage` rather than the extension's if one of them ever imported `urlUtils` or `LocalStorage`. Worth a comment on the guard recording that assumption.

ipdutta (commented):
## Code review summary

Clean, well-scoped port. I checked the build wiring against the Chrome path rather than reading it in isolation, and it's a faithful mirror:

- `exportFirefoxJS` correctly drops `offscreen.js` from the bundle list, and `exportFirefoxSrcFiles` correctly drops `offscreen.html` — so no orphaned offscreen document ships in the XPI.
- The `targetDirHasExportedCommonJs` memo key was added for the new target; missing that would have silently skipped `logManager.js`.
- The `gulp.src`-created-before-`await` pattern in `exportFirefoxJS` looks like a race on `logManager.js`, but `exportChromeJS` does exactly the same thing. Pre-existing, not introduced here.
- esbuild emits `format: "iife"`, so `background.scripts` without `"type": "module"` loads correctly. The asymmetry with Chrome's `service_worker` + `"type": "module"` is right, not an oversight.
- Manifest diff is exactly what it should be: permissions identical apart from dropping `offscreen`, `host_permissions` unchanged at `<all_urls>`, `minimum_chrome_version` dropped, `browser_specific_settings` added. No permission drift in either direction, and the version matches Chrome at `3.11.2`.
- `ClientType.FirefoxExtension` already existed in the enum, so no renumbering.

Keeping `Clipper@OneNote.com` as the gecko id is the right call for upgrade continuity on AMO.

**Status as of `e33a940`: no outstanding items.**

| # | Raised | Location | Outcome |
|---|---|---|---|
| 1 | `getHostname` fallback diverged from the offscreen implementation for opaque-origin schemes | `offscreenCommunicator.ts` | **Fixed** — now builds from `protocol`/`host`, identical to `offscreen.ts` |
| 2 | Unknown-message-type failure mode | `offscreenCommunicator.ts` | **Withdrawn** — my description of the Chrome behaviour was wrong; the original `throw` was defensible. Current `console.warn` + `undefined` is fine |
| 3 | Omits `WebExtension.offscreenUrl` that chrome/edge set | `firefoxExtension.ts` | **Declined, and I agree** — the field is never read, so assigning it would preserve dead state |

Nothing blocking from me.

ipdutta (approved):

src/scripts/communicator/offscreenCommunicator.ts
s-n-d-p: Addressed in e33a940: the fallback now constructs the value from `protocol` and `host`, matching the offscreen implementation for opaque-origin schemes too.

src/scripts/communicator/offscreenCommunicator.ts
s-n-d-p: Addressed in e33a940: unknown message types now warn and return `undefined`, matching the existing offscreen path.

src/scripts/extensions/firefox/firefoxExtension.ts
s-n-d-p: Leaving this unchanged intentionally. `WebExtension.offscreenUrl` is unused, so assigning it in the Firefox entry point would preserve dead state rather than fix behavior. The fallback is currently reached only from extension-owned background/UI contexts; adding a speculative content-script warning would add noise without protecting an active path.

ipdutta:
Verified both fixes at `e33a940` — thanks for the quick turnaround.

`getHostname` now returns `url.protocol + "//" + url.host + "/"`, which is character-for-character the same construction as `getHostname` in `offscreen.ts`. That closes the gap for opaque-origin schemes too: `file:///C:/x.pdf` now yields `"file:///"` on both paths instead of `"null/"` on the fallback. The two implementations are genuinely equivalent now rather than equivalent-in-practice, so the latent dependency on `file://` staying out of the match patterns is gone.

One correction I owe you, on the unknown-message-type thread. I re-checked that path after posting and my original claim was wrong: the offscreen handler returns `false` from `onMessage` *without* calling `sendResponse`, so on Chrome the callback fires with `undefined`, `handleResponse` does `JSON.parse(undefined)`, that throws, and the promise returned by `sendToOffscreenDocument` is never settled. Chrome doesn't cleanly resolve `undefined` there — it hangs. So "matching the existing offscreen path" wasn't an accurate description of what I was asking for, and the `throw` you had originally was defensible.

I deleted that thread once I realised it was wrong, not noticing you'd already addressed it — that was premature on my part and made your reply look orphaned. Apologies for the churn.

No need to revert: `console.warn` + `return undefined` is a fine landing spot, and it's a defensive branch that no current caller can reach since every call site passes an `OffscreenMessageTypes` constant. The Chrome-side hang is pre-existing and out of scope here.

Your reasoning on `WebExtension.offscreenUrl` is right and I'd drop my suggestion — assigning dead state to look symmetric is worse than the asymmetry. Agreed on the content-script point too; it isn't an active path.

Nothing outstanding from me.

ipdutta (approved):

